PR #190, opened 2022-02-02
fix(NavigationBar): Avoid improper Color fallbacks being set on native NavBar elements

kazo0:
## PR Type

- Bugfix

Toolkit uses a `TryGetColorWithOpacity` method from its `ColorHelper` [here](https://github.com/unoplatform/uno.toolkit.ui/blob/39810a2eb6d1e9ad33d672095c24f37fb7ce33c5/src/Uno.Toolkit.UI/Helpers/ColorHelper.cs#L32-L49)

There are certain cases where we do not check if `TryGetColorWithOpacity` returns `false` and would then use the `color` that was set by the method. Problem is that default(Windows.UI.Color) returns a color of `#00000000` and not `null`. The code is structured in a way that it is expecting a `null` as a fallback. This was causing issues such as the UIBarButtonItem's Foreground becoming transparent. We were also not allowing the default iOS UINavigationBar background to display since we would always be changing the background to `#00000000`

If we use the following XAML as an example:
```xml
<utu:NavigationBar Content="Insert Title">
    <utu:NavigationBar.PrimaryCommands>
        <AppBarButton Content="Test" />
        <AppBarButton Content="Test 2" />
    </utu:NavigationBar.PrimaryCommands>
</utu:NavigationBar>
```

We would get something like this:
image

## What is the new behavior?

Now it should look like this:
image

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationBarRenderer.iOS.cs
+++ b/src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationBarRenderer.iOS.cs
@@ -81,68 +81,85 @@ namespace Uno.Toolkit.UI
 			var appearance = new UINavigationBarAppearance();
 
 			// Background
-			ColorHelper.TryGetColorWithOpacity(Element.Background, out var backgroundColor);
-			switch (backgroundColor)
+			if (ColorHelper.TryGetColorWithOpacity(Element.Background, out var backgroundColor))
 			{
-				case { } opaqueColor when opaqueColor.A == byte.MaxValue:
-					if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
-					{
-						appearance.ConfigureWithOpaqueBackground();
-						appearance.BackgroundColor = opaqueColor;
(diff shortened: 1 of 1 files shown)


src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationBarRenderer.iOS.cs
```diff
@@ -251,6 +274,10 @@ private void ApplyVisibility()
 			}
 		}
 
+		private Windows.UI.Color? GetMainCommandForeground()
+		{
+
+		}
```
jeromelaban, line 280: Empty on purpose?

jeromelaban (approved):

src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationBarRenderer.iOS.cs
kazo0: Oops, deleted

PR #197, opened 2022-03-07
fix: Re-work resource names to avoid issues with xaml merge

kazo0:
closes https://github.com/unoplatform/uno/issues/8224
closes https://github.com/unoplatform/uno/issues/8230
closes https://github.com/unoplatform/uno/issues/8228
closes https://github.com/unoplatform/uno/issues/8224

## PR Type

- Bugfix

Prior to introducing the xaml merge task, each Resource Dict in Toolkit.Material and Toolkit.Cupertino were re-defining certiain static resources using the same key but in different file. Now with the xaml merge, we need to use more specific resource keys for each control


jeromelaban:
It feels like we should be changing the xaml merge tool to detect these overrides, as it hides issues.

jeromelaban (approved):

Issue #199, opened 2022-03-10
System.InvalidOperationException: Cannot locate resource from 'ms-appx:///Uno.Toolkit.UI.Material/Styles/Application/Colors.xaml'


anpin:
I referenced Uno.Toolkit.WinUI.Material 1.3.0 and followed the guide to define resources, but it gives me following exception on startup 
## Current behavior
Throw this on startup
```[mono-rt] [ERROR] FATAL UNHANDLED EXCEPTION: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
[mono-rt]  ---> System.InvalidOperationException: Cannot locate resource from 'ms-appx:///Uno.Toolkit.UI.Material/Styles/Application/Colors.xaml'
[mono-rt]    at Uno.UI.ResourceResolver.RetrieveDictionaryForSource(String source, String currentAbsolutePath) in C:\a\1\s\src\Uno.UI\UI\Xaml\ResourceResolver.cs:line 469
[mono-rt]    at Uno.Toolkit.UI.Material.GlobalStaticResources.ResourceDictionarySingleton__1.get_mergedpages_winui_6cf9cf3955d7d3d75b0c26a59d8c9209_ResourceDictionary() in D:\a\1\s\src\library\Uno.Toolkit.Material\obj\Uno.Toolkit.WinUI.Material\Release\monoandroid11.0\g\XamlCodeGenerator\mergedpages_winui_6cf9cf3955d7d3d75b0c26a59d8c9209.g.cs:line 1216
[mono-rt]    at Uno.Toolkit.UI.Material.GlobalStaticResources.get_mergedpages_winui_6cf9cf3955d7d3d75b0c26a59d8c9209_ResourceDictionary() in D:\a\1\s\src\library\Uno.Toolkit.Material\obj\Uno.Toolkit.WinUI.Material\Release\monoandroid11.0\g\XamlCodeGenerator\mergedpages_winui_6cf9cf3955d7d3d75b0c26a59d8c9209.g.cs:line 1836
[mono-rt]    at Uno.Toolkit.UI.Material.GlobalStaticResources.<>c.<RegisterResourceDictionariesBySource>b__9_1() in D:\a\1\s\src\library\Uno.Toolkit.Material\obj\Uno.Toolkit.WinUI.Material\Release\monoandroid11.0\g\XamlCodeGenerator\GlobalStaticResources.g.cs:line 97
[...]

PR #200, opened 2022-03-15
fix!: drop support for MaterialToolkitColors

kazo0:
BREAKING CHANGE: MaterialToolkitColors no longer exists

fixes https://github.com/unoplatform/uno.toolkit.ui/issues/199

## PR Type

- Bugfix
- Documentation content changes

WinUI heads would crash at runtime

## What is the new behavior?

WinUI heads do not crash


kazo0:
@jeromelaban, for creating tests for this, would we just be able to run the current UI Tests on both the iOS/Android UWP and WinUI heads?
Just to make sure that WinUI project are compiling and not crashing at runtime?

jeromelaban:
For what I can see, UI tests are not setup for this repo, so we won't catch much yet. We should be able to do the same kind of updates that the Uno.Themes repro does.

nickrandolph (approved):

PR #204, opened 2022-03-30
feat: Add .NET 6 mobile support

jeromelaban:


kazo0 (approved):

Xiaoy312 (approved):

kazo0 (commented):
Do we need to add `MonoAndroid12.0` to the TargetFrameworks?

jeromelaban:
> Do we need to add `MonoAndroid12.0` to the TargetFrameworks?

We don't, as this library does not use Android 12 features explicitly.

PR #259, opened 2022-06-08
fix(NavigationBar): improperly setting foreground/background to appbarbutton

kazo0:
Fix error that was setting the default AppBarButton within NavigationBar to the improper foreground/background

Adjust Height of XAML-based NavBar to 64


agneszitte-nventive (approved):

agneszitte-nventive:
should fix the Height for the NavBar please

agneszitte-nventive (commented):
should fix the Height for the NavBar please at the same time @kazo0

carldebilly (approved):

PR #324, opened 2022-09-16
fix(SafeArea): Ensure the root of the XAML tree is used to compute relative bounds of SafeArea control/owner

kazo0:
closes https://github.com/unoplatform/nventive-private/issues/412

## PR Type

- Bugfix

SafeArea calculates improper offsets when the control is within a native iOS Modal window. We were using `TransformToVisual` relative to `Windows.Current.Content` which results in an offset of X=16, Y=47 since the Window.Content (white background) is no longer at position 0,0 while a modal (red background) is being displayed

MicrosoftTeams-image (2)

## What is the new behavior?

Use `TransformToVisual(null)` to ensure that the root of the XAML tree is used for the transformation

### DIFF
--- a/src/Uno.Toolkit.UI/Controls/SafeArea/SafeArea.cs
+++ b/src/Uno.Toolkit.UI/Controls/SafeArea/SafeArea.cs
@@ -354,7 +354,9 @@ namespace Uno.Toolkit.UI
 					// If the owner view is scrollable, the visibility of interest is that of the scroll viewport.
 					var fixedControl = scrollAncestor ?? Owner;
 
-					var controlBounds = GetRelativeBounds(fixedControl, XamlWindow.Current.Content);
+					// Using relativeTo: null instead of Window.Current.Content since there are cases when the current UIElement
+					// may be outside the bounds of the current Window content, for example, when the element is hosted in a modal window.
+					var controlBounds = GetRelativeBounds(fixedControl, relativeTo: null);
 
 					visibilityPadding = CalculateVisibilityInsets(OffsetVisibleBounds, controlBounds);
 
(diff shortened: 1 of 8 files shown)


src/Uno.Toolkit.UI/Controls/SafeArea/SafeArea.cs
```diff
@@ -354,7 +354,9 @@ private void UpdateInsets()
 					// If the owner view is scrollable, the visibility of interest is that of the scroll viewport.
 					var fixedControl = scrollAncestor ?? Owner;
 
-					var controlBounds = GetRelativeBounds(fixedControl, XamlWindow.Current.Content);
+					// Using relativeTo: null instead of Window.Current.Content since there are cases when the current UIElement
```
jeromelaban, line 357: Is there a way to UI Test this ?

kazo0: Added!

samples/Uno.Toolkit.Samples/Uno.Toolkit.Samples.Droid/Uno.Toolkit.Samples.Droid.csproj
```diff
@@ -16,7 +16,6 @@
     <AndroidResgenFile>Resources\Resource.Designer.cs</AndroidResgenFile>
     <AndroidUseAapt2>true</AndroidUseAapt2>
     <GenerateSerializationAssemblies>Off</GenerateSerializationAssemblies>
-    <AndroidUseLatestPlatformSdk>False</AndroidUseLatestPlatformSdk>
```
jeromelaban, line 19: Why did you need this ?

jeromelaban (approved):

PR #372, opened 2022-10-17
ci: include net6 builds

kazo0:
Avoid accidentally disabling net6.0 target frameworks


build/workflow/pipeline.yml
```diff
@@ -55,12 +55,10 @@ jobs:
       Samples:
         ApplicationPlatform: Samples
         ArtifactName: Samples
-        DisableNet6MobileTargets: true
```
jeromelaban, line 58: Why do you remove those? That is unless samples also now build with net6, but I don't think they do.

jeromelaban (approved):

PR #393, opened 2022-11-03
feat: Enable XAML trimming support

jeromelaban:


kazo0 (approved):

Xiaoy312 (commented):
I assume in a way this is trading build time for lib size. Shouldn't we only enable this for non-debug build?

jeromelaban:
Enabling this on a library only includes a set of generated classes, it does not impact the build time (it's only at the app level that it does, and it's not enabled here).

PR #408, opened 2022-11-23
perf: Improve IL trimming

jeromelaban:


src/Uno.Toolkit.UI/Behaviors/DrawerControlBehavior.cs
```diff
@@ -33,73 +34,87 @@ public static class DrawerControlBehavior
...
 
-		public static DependencyProperty OpenDirectionProperty { get; } = DependencyProperty.RegisterAttached(
+		public static DependencyProperty OpenDirectionProperty { [DynamicDependency(nameof(GetOpenDirection))] get; } = DependencyProperty.RegisterAttached(
 			"OpenDirection",
 			typeof(DrawerOpenDirection),
 			typeof(DrawerControlBehavior),
 			new PropertyMetadata(default(DrawerOpenDirection)));
 
+		[DynamicDependency(nameof(SetOpenDirection))]
 		public static DrawerOpenDirection GetOpenDirection(DependencyObject obj) => (DrawerOpenDirection)obj.GetValue(OpenDirectionProperty);
```
kazo0, line 59: Are the Sets and Gets supposed to be flipped? 

Like DynamicDependency("SetOpenDirection") is on the GetOpenDirection method.

kazo0: Oh I see, just looked up DynamicDependency. Nevermind

kazo0 (approved):

src/Uno.Toolkit.UI/Behaviors/DrawerControlBehavior.cs
jeromelaban: They are! The idea is that if you use the Set, the Get should stay (even if not referenced directly); if you use the Get, the set should stay; and if you use the DP auto property, both the get and set should say (transitively). We'll modify the uno generator (and probably extract it), so it does this automatically somehow.

Xiaoy312 (approved):

PR #467, opened 2023-02-02
fix: Adjust measure/arrange for TabBarListPanel

kazo0:
closes https://github.com/unoplatform/uno.chefs/issues/249


agneszitte-nventive (commented):
@kazo0 Can we maybe add some tests for TabBar in order to prevent in the future similar issues?

agneszitte-nventive (approved):

PR #517, opened 2023-03-20
fix: avoid winappsdk crash

kazo0:


src/library/Uno.Toolkit.Material/Styles/Controls/v2/TabBar.xaml
```diff
@@ -43,6 +43,18 @@
 	<Thickness x:Key="MaterialNavigationTabBarItemLargeBadgePadding">4,0</Thickness>
 	<CornerRadius x:Key="MaterialNavigationTabBarItemLargeBadgeCornerRadius">8</CornerRadius>
 
+	<!-- WORKAROUND: ControlExtensions must be explicitly referenced to avoid a crash on WinAppSDK apps -->
```
jeromelaban, line 46:
```suggestion
	<!-- WORKAROUND: ControlExtensions must be explicitly referenced as an attribute to avoid a crash on WinAppSDK apps, and not just as a a setter property -->
```

jeromelaban:
@mergifyio backport release/stable/2.5

nickrandolph:
@kazo0 is this a known issue on the WinAppSdk? If so, can we add a reference to the issue number to this PR and if not, I guess we should lodge an issue

PR #556, opened 2023-04-26
ci: add support for canaries

kazo0:


jeromelaban (approved):

Xiaoy312 (approved):

kazo0:
@jeromelaban looks like we'd need new .mobileprovision files for these new Bundle IDs. I changed them to differentiate between winui and non-winui as well as the new canary bundle ids. I don't think I have permissions to create new certs/provisioning profiles

kazo0:
@jeromelaban this would be for the new Toolkit Canary apps though, I can re-use the ones that are for Chefs? I thought they'd be tied to a specific bundle id

agneszitte (approved):

agneszitte:
> Ah that's toolkit, yes I'll fix that

agneszitte:
> Ah that's toolkit, yes I'll fix that :)

@kazo0 you should be able to use:
image

samples/Uno.Toolkit.WinUI.Samples/Uno.Toolkit.WinUI.Samples.macOS/Info.plist
```diff
@@ -14,7 +14,7 @@
 	<string>10.11</string>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
-	<key>CFBundleInfoDictionaryVersion</key>
+<key>CFBundleInfoDictionaryVersion</key>
```
agneszitte, line 17:
```suggestion
	<key>CFBundleInfoDictionaryVersion</key>
```

Issue #586, opened 2023-05-30
[iOS][Android] AppBarButton breaks NavigationBar if Icon specified


mikeirvingweb:
## Current behavior

Navigation Bar does not render (at all) on Mobile, if an Icon is used

error in Output - `System.InvalidOperationException: The Dependency Property [UriSource] is owned by [Microsoft.UI.Xaml.Controls.BitmapIcon] and cannot be used on [Microsoft.UI.Xaml.Controls.SymbolIcon]`

## Expected behavior

Navigation Bar renders, and the button has the desired icon, like on Windows

## How to reproduce it (as minimally and precisely as possible)

`<utu:NavigationBar Content="Macclesfield Hedgehogs">`
`	<utu:NavigationBar.PrimaryCommands>`
`		<AppBarButton Label="Refresh" Icon="Refresh" />`
`	</utu:NavigationBar.PrimaryCommands>`
`</utu:NavigationBar>`

Affected platform(s):

- [X] Android
- [X] iOS

kazo0:
Thanks for the report @mikeirvingweb!

Unfortunately only BitmapIcons can be supported by the NavigationBar for Mobile targets at this time as the underlying native components do no support Font/Path/Symbol Icons. See [here](https://platform.uno/docs/articles/external/uno.toolkit.ui/doc/controls/NavigationBar.html#how-can-i-use-a-path-for-the-appbarbutton-icon)

Some investigation still needs to be done to see if this would be possible in the future but for right now we only have support for `BitmapIcon`.

mikeirvingweb:
Yeah. A pain though. It's nice to be able to use Symbol Icons elsewhere, but painful to have to make bitmaps, or use just text, here.

Yes, I feel it should not break the navbar altogether, and if the error made the cause obvious it would save devs a lot of time

mikeirvingweb:
By 'not breaking' the nav bar, it should just default to the `Label`, if one has been provided.

PR #589, opened 2023-06-05
fix: NavBar Icon fixes

kazo0:
related https://github.com/unoplatform/nventive-private/issues/487
related https://github.com/unoplatform/uno/issues/8996
fixes https://github.com/unoplatform/uno.toolkit.ui/issues/586

Fix error when using non-supported appbarbutton iconelement

Setting the Foreground property of a BitmapIcon has no effect when being used within the native renderers for CommandBar on iOS/Android

## What is the new behavior?

Foreground on BitmapIcon is respected and takes precedence over the Foreground of the AppBarButton

### DIFF
--- a/src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationAppBarButtonRenderer.Android.cs
+++ b/src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationAppBarButtonRenderer.Android.cs
@@ -125,25 +125,27 @@ namespace Uno.Toolkit.UI
 
 			bool TrySetIcon()
 			{
-				if (element.Icon is { } icon)
+				if (element.Icon is not { } icon)
 				{
-					if (icon is BitmapIcon bitmap)
-					{
-						if (bitmap.UriSource is { } uriSource)
-						{
-							native.NavigationIcon = DrawableHelper.FromUri(uriSource);
-							if (hasIconColor)
-							{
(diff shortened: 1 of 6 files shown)


src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationAppBarButtonRenderer.Android.cs
```diff
@@ -75,38 +75,23 @@ protected override IEnumerable<IDisposable> Initialize()
 		protected override void Render()
 		{
 			var native = Native;
-			var element = Element;
+			var element = Element ?? throw new InvalidOperationException("Element is null.");
```
ajpinedam, line 78: This is neat!

src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationAppBarButtonRenderer.Android.cs
```diff
@@ -137,6 +122,32 @@ protected override void Render()
 				native.NavigationIcon = null;
 				native.NavigationContentDescription = null;
 			}
+
+			bool TrySetIcon()
+			{
+				if (element.Icon is { } icon)
```
ajpinedam, line 128: Wouldn't early guard avoid this deep indentation?

```chsarp
if (element.Icon is not { } icon) 
{
    return false;
}
```

ajpinedam (approved):

PR #622, opened 2023-06-21
fix: Properly resubscribe to nested prop changes

### LINKED ISSUE #381: [NavigationBar][iOS][Android] Cannot use Converters for properties on NavigationBar and AppBarButton on mobile
See [this issue](https://github.com/unoplatform/uno.todo/issues/231)

So it appears that we cannot use a Converter when setting the Background of the NavigationBar. For some reason, it breaks the nested registration of the PropertyChanged callback here: https://github.com/unoplatform/uno.toolkit.ui/blob/650812afeed82afc5b330a28b50ba97625331446/src/Uno.Toolkit.UI/Controls/NavigationBar/NavigationBarRenderer.Android.cs#L151-L153

kazo0:
fixes #381 

## PR Type

- Bugfix

When a property would change at runtime, the logic to re-register to the nested property callback was incorrectly using the `sender` instead of the actual owner of the sub-properties

### DIFF
--- a/src/Uno.Toolkit.RuntimeTests/Tests/NavigationBarTests.cs
+++ b/src/Uno.Toolkit.RuntimeTests/Tests/NavigationBarTests.cs
@@ -22,8 +22,10 @@ using Microsoft.UI.Xaml.Controls.Primitives;
 using Microsoft.UI.Xaml.Media;
 using Microsoft.UI.Xaml.Navigation;
 using Microsoft.UI.Xaml;
+using Microsoft.UI;
 #else
 using Windows.UI.Xaml.Controls;
+using Windows.UI;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls.Primitives;
 using Windows.UI.Xaml.Markup;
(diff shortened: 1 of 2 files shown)


src/Uno.Toolkit.RuntimeTests/Tests/NavigationBarTests.cs
```diff
@@ -391,17 +429,18 @@ public NavBarTestPage()
 #endif
 	}
 
-#if __IOS__
+#if __IOS__ || __ANDROID__
 	public static class NavigationBarTestHelper
 	{
+#if __IOS__
 		public static UINavigationBar? GetNativeNavBar(this NavigationBar? navBar) => navBar
```
rajamatt, line 436: I think it would be good to have the following from GetNativeNavBar indented the same way as GetNativeNavItem

`?.TryGetRenderer<NavigationBar, NavigationBarRenderer>()` 
`?.Native;`

rajamatt (approved):

agneszitte (approved):

agneszitte:
https://github.com/Mergifyio backport legacy/3x

Issue #623, opened 2023-06-23
Exception in AppBarButton when it is inside a NavigationBar in Android


Suriman:
### Current behavior

The AppBarButton works correctly in WinUI but:
	- Does not support SVG images in Wasm and Android.
	- When it is included in a NavigationBar it raises exception if the content is different of a BitmapImage in Android platform.

### Expected behavior

The AppBarButton supports all functionality on all platforms.

### How to reproduce it (as minimally and precisely as possible)

Open the attached project and compile it. 
Execute the Android project and the NavigationBar fails to load because the AppBarButton image load throws an exception, if all the AppBarButton are commented except the one containing the BitmapImage it loads correctly.

Code that reproduces the error in this file:
[UnoAppBarButtonAndroid.zip](https://github.com/unoplatform/uno/files/11850172/UnoAppBarButtonAndroid.zip)

### Workaround

No workaround.

### Works on UWP/WinUI

Yes

### Environment

Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia

### NuGet package version(s)

Uno.WinUI 4.9.20

### Affected platforms

WebAssembly, Android

### IDE

Visual Studio 2022

### IDE version

17.6.3

kazo0:
@Suriman Looking into this now, I am able to see the FontIcons on WASM, is there another issue on WASM that you are experiencing?
![image](https://github.com/unoplatform/uno.toolkit.ui/assets/4793020/053c2ad8-c9ff-4436-9f7a-9380c70fbfce)

Also, I have updated the package versions used in the sample app to the latest dev bits and the NavigationBar will now render, although the AppBarButtons that are not using a `BitmapIcon` will not display as this is still a limitation of the native bar being used on Android and iOS:
![image](https://github.com/unoplatform/uno.toolkit.ui/assets/4793020/f46ca4a9-d9a0-44ab-9eaf-28862314e2c8)


Attached is the up-to-date repro app
[AddInfo.zip](https://github.com/unoplatform/uno.toolkit.ui/files/11872420/AddInfo.zip)

ADD-Eugenio-Lopez:
Yes, we know the limitation of the documentation that on Android it only works with BitmapImage, that's why we have commented it with Sasha and the issue has been opened as feature enhancement.

We have updated the package versions used in the sample app to the latest dev bits.
In WinUI we can see all the image sources:
[...]

PR #631, opened 2023-06-30
feat: Support all IconElements for native NavBar

kazo0:
fixes https://github.com/unoplatform/uno.toolkit.ui/issues/623

## PR Type

- Feature

Only BitmapIcons are supported in the Android and iOS native NavigationBar renderers

## What is the new behavior?

All IconElements are supported

### DIFF
--- a/src/Uno.Toolkit.UI/Controls/NavigationBar/AppBarButtonRenderer.Android.cs
+++ b/src/Uno.Toolkit.UI/Controls/NavigationBar/AppBarButtonRenderer.Android.cs
@@ -136,6 +136,37 @@ namespace Uno.Toolkit.UI
 			native.SetActionView(null);
 			native.SetIcon(null);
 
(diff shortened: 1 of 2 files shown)


Youssef1313 (commented):
Isn't that exactly a copy of what we have in uno? Could the code be re-used instead of duplicated?

Youssef1313 (approved):

kazo0:
> Isn't that exactly a copy of what we have in uno? Could the code be re-used instead of duplicated? (InternalsVisibleTo for example)

@Youssef1313 yes, I wasn't allowed to make InternalsVisibleTo the Toolkit :(

agneszitte (approved):

agneszitte (commented):
Can we add some tests for this ?

kazo0:
replaced by https://github.com/unoplatform/uno.toolkit.ui/pull/1045

PR #701, opened 2023-08-07
Small improvements

pictos:
I'm not using the template, because I added two small fixes on the code.

1. Change the `SKShadow` to be a `readonly record struct`, I believe we want it to be immutable, and `record struct` isn't immutable, we to add the `readonly` modifier;
2. Use `Array.Empty<T>` to avoid allocations.

### DIFF
--- a/src/Uno.Toolkit.Skia.WinUI/Controls/Shadows/ShadowContainer.Paint.cs
+++ b/src/Uno.Toolkit.Skia.WinUI/Controls/Shadows/ShadowContainer.Paint.cs
@@ -78,7 +78,7 @@ public partial class ShadowContainer
 		}
 	}
 
-	private ShadowInfos[] _shadowInfoArray = new[] { ShadowInfos.Empty };
+	private ShadowInfos[] _shadowInfoArray = Array.Empty<ShadowInfos>();
 	private float _currentPixelRatio;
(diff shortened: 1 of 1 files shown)


src/Uno.Toolkit.Skia.WinUI/Controls/Shadows/ShadowContainer.Paint.cs
```diff
@@ -78,7 +78,7 @@ public static SKShadow From(Shadow shadow, float width, float height, float corn
 		}
 	}
 
-	private ShadowInfos[] _shadowInfoArray = new[] { ShadowInfos.Empty };
+	private ShadowInfos[] _shadowInfoArray = Array.Empty<ShadowInfos>();
```
Youssef1313, line 81: This changes the meaning of the code. Is it intentional?

pictos: Yes, it's. If you look into the code, this a `private` field and is used just to hold the `oldShadowCollection`, basically to see if something changed and should invalidate something. So, AFAIK, it's a safe change

Xiaoy312 (approved):

agneszitte:
@Mergifyio backport legacy/3x

Issue #874, opened 2023-10-13
[iOS][Dynamic Island] NavigationBar extra bottom spacing 


kazo0:
On iPhones with the Dynamic Island (14+), there is extra space below the NavigationBar, likely caused by the SafeArea.

[SafeAreaTest2.zip](https://github.com/unoplatform/uno.toolkit.ui/files/12896836/SafeAreaTest2.zip)

iPhone 14 Pro Max (16.4)
![image](https://github.com/unoplatform/uno.toolkit.ui/assets/4793020/06a2ad4c-9b42-4228-af53-c45565539e5c)

iPhone 13 Pro (16.4)
![image](https://github.com/unoplatform/uno.toolkit.ui/assets/4793020/51102fd0-7d4b-4762-a7e5-692fecafbb48)

kazo0:
Note, this is occurring only when SafeArea insets are set on the Style of the NavigationBar. If I remove the Setter for `utu:SafeArea.Insets` and instead set it directly on the NavigationBar instance in the sample XAML, no line appears

PR #876, opened 2023-10-14
fix: templatebind background to ios navbar template

kazo0:
fixes https://github.com/unoplatform/uno.toolkit.ui/issues/874

## PR Type

- Bugfix

image

## What is the new behavior?

image


Xiaoy312 (commented):
this is just a visual bug with the sample, and not the logic/calculation of safe area?

kazo0:
It's actually something in UIKit, the safe area reported by the os is larger than the native bars that it draws and we don't have control over the height of the UINavigationBar

agneszitte (approved):

PR #912, opened 2023-11-13
feat: ResponsiveView

eriklimakc:
GitHub Issue (If applicable): #

## PR Type

## PR Checklist

	- [ ] UWP
	- [ ] WinUI
	- [ ] iOS
	- [ ] Android
	- [ ] WASM
	- [ ] MacOS

## Other information


eriklimakc:
# To Discuss:

## 1. How we handle Template Content
What should be done when, for example, Narrow VisualState is triggered but NarrowContent is `null`? Should the DefaultContent be displayed instead? What would be the order for other states like Wide/ExtraWide?

https://github.com/unoplatform/uno.toolkit.ui/blob/67ff5a47b7c4abf2e816f1673639efd17287a5e3/src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.cs#L97-L114

## 2. How to make the responsive settings global and overridable?
### 2.1 What should be the default setting for ExtraNarrow, Narrow, Default, etc... ?
https://github.com/unoplatform/uno.toolkit.ui/blob/67ff5a47b7c4abf2e816f1673639efd17287a5e3/src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.xaml#L7-L23

## 3. Properties naming
Would it make more sense if the we named them, for example, `ExtraNarrowTemplate` instead of `ExtraNarrowContent`, or `ExtraNarrowTemplateContent`?
https://github.com/unoplatform/uno.toolkit.ui/blob/67ff5a47b7c4abf2e816f1673639efd17287a5e3/src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.cs#L14

cc @Xiaoy312 @kazo0

src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.cs
```diff
@@ -0,0 +1,145 @@
...
+		SetContent(contentToSet);
+	}
+
+	private void SetContent(DataTemplate? contentToSet)
+	{
+		if (contentToSet is not null)
+		{
+			Content = contentToSet.LoadContent() as UIElement;
+		}
+	}
```
Xiaoy312, line 144: just set the content-template directly
and also, no need a null guard, we actually want this to clear if no match
```suggestion
	public void OnSizeChanged(Size size, ResponsiveLayout layout)
	{
		ContentTemplate = GetValueForSize(size, ResponsiveLayout ?? layout);
	}
```

src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.cs
Xiaoy312, line 1: beside, on-loaded and size-changed, you should also re-calculate the content when any of 5 {narrow,wide,...} DPs changes (but dont update it if `this.IsLoaded` is false, since they could be hit when this control is being parsed)

src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.cs
```diff
@@ -0,0 +1,145 @@
...
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+#endif
+
+namespace Uno.Toolkit.UI;
+
+public partial class ResponsiveView : ContentControl
+#if !WINDOWS_UWP
+	, IResponsiveCallback
+#endif
```
Xiaoy312, line 17:
```suggestion
public partial class ResponsiveView : ContentControl, IResponsiveCallback
```

src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.cs
eriklimakc: Done in my next commits

src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.cs
eriklimakc: Done in my next commits

src/Uno.Toolkit.UI/Controls/ResponsiveView/ResponsiveView.cs
eriklimakc: Done in my next commits

eriklimakc:
@kazo0 and @Xiaoy312 , please note that my latest changes include:

```
#if WINDOWS || WINDOWS_UWP
	Content = dataTemplate?.LoadContent() as UIElement;
#else
        ContentTemplate = dataTemplate;
#endif
```

The reason behind this is that using `ContentTemplate` on non-windows platforms gives us a better experience when resizing the screen, see below:
### ContentTemplate
ContentTemplate

### Content
Content

\* Using `ContentTemplate` on Windows doesn't work.

Xiaoy312:
It shouldnt flicker this much on windows, with the Content. You need to cache the last resolved content-template or winning size, and compare this value on size-changed, and only update when the new ones are different than previous.
do not compare the result of ContentTemplate.Load() that is guaranteed to be different.

doc/controls/ResponsiveView.md
```diff
@@ -0,0 +1,131 @@
...
+| Narrow     | double | Default value is 300.  |
+| Normal     | double | Default value is 600.  |
+| Wide       | double | Default value is 800.  |
+| Widest     | double | Default value is 1080. |
+
+### Usage
+
+```xml
+xmlns:utu="using:Uno.Toolkit.UI"
+xmlns:hlp="using:Uno.Toolkit.UI.Helpers"
```
kazo0, line 93: How come ResponsiveLayout is in a different namespace? We could have everything inside of Uno.Toolkit.UI that should be used publicly, no?

src/Uno.Toolkit.UI/Helpers/ResponsiveHelper.cs
```diff
@@ -0,0 +1,159 @@
...
+using Uno.Disposables;
+
+#if IS_WINUI
+using Microsoft.UI.Xaml;
+#else
+using Windows.UI.Xaml;
+using Windows.UI.Core;
+#endif
+
+namespace Uno.Toolkit.UI.Helpers;
```
kazo0, line 13: Keep it as Uno.Toolkit.UI namespace

src/Uno.Toolkit.UI/Helpers/ResponsiveHelper.cs
```diff
@@ -0,0 +1,159 @@
...
+		var wr = new WeakReference(host);
+		_references.Add(wr);
+	}
+
+	internal static IDisposable UsingDebuggableInstance()
+	{
+		UseDebuggableInstance = true;
+
+		return Disposable.Create(() => UseDebuggableInstance = false);
+	}
```
kazo0, line 158: Should this be within an #if DEBUG? Just thinking out loud, not sure if it makes sense or not

cc: @Xiaoy312

samples/Uno.Toolkit.Samples/Uno.Toolkit.Samples.Shared/App.xaml.cs
```diff
@@ -98,6 +99,8 @@ protected override async void OnLaunched(XamlLaunchActivatedEventArgs e)
 #else
 			_window = XamlWindow.Current;
 #endif
+			var helper = ResponsiveHelper.GetForCurrentView();
+			helper.HookupEvent(_window);
```
kazo0, line 103: Is this needed just for the runtime tests? This isn't needed for people to use the control in general, right?

doc/controls/ResponsiveView.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Properties
+| Property          | Type             | Description                                             |
+| ----------------- | ---------------- | ------------------------------------------------------- |
+| NarrowestTemplate | DataTemplate     | Template to be displayed on the narrowest screen size.  |
+| NarrowTemplate    | DataTemplate     | Template to be displayed on a narrow screen size.       |
+| NormalTemplate    | DataTemplate     | Template to be displayed on a normal screen size.       |
+| WideTemplate      | DataTemplate     | Template to be displayed on a wide screen size.         |
+| WidestTemplate    | DataTemplate     | Template to be displayed on the widest screen size.     |
+| ResponsiveLayout  | ResponsiveLayout | Overrides the screen size threshold/breakpoints.        |
```
kazo0, line 41: This property can be renamed to `Layout` to match the ResponsiveExtension properties

src/Uno.Toolkit.UI/Helpers/ResponsiveHelper.cs
Xiaoy312: this is a backdoor used by runtime tests
having DEBUG will make them inaccessible

samples/Uno.Toolkit.Samples/Uno.Toolkit.Samples.Shared/App.xaml.cs
Xiaoy312: HookupEvent is needed to receive SizeChanged event

doc/controls/ResponsiveView.md
Xiaoy312: i think there was a name clash with one of method of the same name, which is why we added the Responsive- prefix

doc/controls/ResponsiveView.md
```diff
@@ -0,0 +1,131 @@
...
+
+### XAML
+```xml
+xmlns:utu="using:Uno.Toolkit.UI"
+...
+
+<utu:ResponsiveView>
+     <!-- Contents -->
+</utu:ResponsiveView>
+```
```
Xiaoy312, line 23: this section is completely useless...
remove

doc/controls/ResponsiveView.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Inheritance 
+Object &#8594; DependencyObject &#8594; UIElement &#8594; FrameworkElement &#8594; Control &#8594; ContentControl
+
+### Constructors
+| Constructor      | Description                                               |
+|------------------|-----------------------------------------------------------|
+| ResponsiveView() | Initializes a new instance of the `ResponsiveView` class. |
```
Xiaoy312, line 31: remove; completely useless

doc/controls/ResponsiveView.md
Xiaoy312, line 1: need a section here to explain the resolution logics between current screen size vs the defined templates
especially when some templates are omitted
since you cant expect them to always define all 5 of them
when realistically Narrow+Wide is probably the common setup

Xiaoy312: > var match = defs.FirstOrDefault(y => y?.MinWidth >= size.Width) ?? defs.LastOrDefault();

smallest defined templated that fit the width, or the largest defined templated

Xiaoy312, line 1: the layout is a mess
```
# ResponsiveView
## Summary // why is almost everything nested under summary?
### C#
### XAML
### Inheritance 
### Constructors
### Properties
### Usage
## ResponsiveLayout 
### Properties
### Usage
```
rearrange this page to be like:
```
# ResponsiveView
## Summary // i would also argue to put this section directly under #ResponsiveView as text content without its own header
## Inheritance // questionable if needed or not
## Properties
### ResponsiveLayout
## Usage // use cases: with templates, with custom breakpoints(ResponsiveLayout)
```

Xiaoy312 (approved):

samples/Uno.Toolkit.Samples/Uno.Toolkit.Samples.Shared/App.xaml.cs
eriklimakc: I guess this should be on the docs as well, right @Xiaoy312 ?

src/Uno.Toolkit.UI/Helpers/ResponsiveHelper.cs
eriklimakc: Done in my next commit

doc/controls/ResponsiveView.md
eriklimakc: Done in my next commit

doc/controls/ResponsiveView.md
eriklimakc: Done in my next commit

Issue #382, opened 2022-10-27
[SafeArea][WinUI] SafeArea should be using XamlRoot and not Window.Current


kazo0:
In spots like [here](https://github.com/unoplatform/uno.toolkit.ui/blob/0d2a3606cd1398e3bf78589ac59229f39adb618a/src/Uno.Toolkit.UI/Controls/SafeArea/SafeArea.cs#L195) and multiple other places, we should be using the XamlRoot [Uno](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/XamlRoot.cs) [Windows](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.xamlroot?view=winrt-22621) since Window.Current returns null on WinUI.

agneszitte:
@MartinZikmund is this something to consider for multi-window support ?

MartinZikmund:
yes, very likely, on WinAppSDK and Uno.WinUI apps `Window.Current` would not work

PR #1030, opened 2024-02-02
fix: add proper null checks for Window.Current

kazo0:
fixing [canary failures](https://dev.azure.com/uno-platform/Uno%20Platform/_build/results?buildId=111044&view=results)

### DIFF
--- a/src/Uno.Toolkit.UI/Controls/SafeArea/SafeArea.cs
+++ b/src/Uno.Toolkit.UI/Controls/SafeArea/SafeArea.cs
@@ -237,7 +237,7 @@ namespace Uno.Toolkit.UI
 				if (inputRect.Top != 0 && inputRect.Top < visibleBounds.Bottom)
 				{
 
-					var windowBottom = XamlWindow.Current.Bounds.Bottom;
+					var windowBottom = XamlWindow.Current?.Bounds.Bottom ?? 0d;
 #if __ANDROID__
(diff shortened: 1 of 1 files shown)


src/Uno.Toolkit.UI/Controls/SafeArea/SafeArea.cs
```diff
@@ -237,7 +237,7 @@ private static Rect GetVisibleBounds(Rect? safeAreaOverride = null, bool withSof
 				if (inputRect.Top != 0 && inputRect.Top < visibleBounds.Bottom)
 				{
 
-					var windowBottom = XamlWindow.Current.Bounds.Bottom;
+					var windowBottom = XamlWindow.Current?.Bounds.Bottom ?? 0d;
```
nickrandolph, line 240: Does SafeArea correctly handle when XamlWindow.Current is null? And is this only for WinUI, or are there other scenarios where it might be null

MartinZikmund: Ideally it should use `XamlRoot` instead, as we will ultimately make `Current` null across Uno.WinUI (kept it non null to avoid breaking changes for now)

kazo0: Yeah there is still work that needs to be done to make it work with XamlRoot: https://github.com/unoplatform/uno.toolkit.ui/issues/382

This was just a stop-gap to fix the canaries

PR #1048, opened 2024-02-20
chore: remove nested test classes

kazo0:


src/Uno.Toolkit.RuntimeTests/Tests/NavigationBarTests.cs
Youssef1313, line 1: I would prefer updating Uno to a version that has this issue fixed.

Youssef1313: The fix was https://github.com/unoplatform/uno/pull/14324 (5.1.0-dev.443) and was backported to 5.0

Youssef1313: Looks like toolkit is using 5.0.19 which is quite old. Any blocker for updating to 5.0.154?

PR #1068, opened 2024-03-08
feat: add support for iselectioninfo to itemsrepeater extensions

kazo0:


src/Uno.Toolkit.UI/Behaviors/ItemsRepeaterExtensions.cs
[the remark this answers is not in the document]
kazo0: I'm not sure I understand why any of that would be necessary? Shouldn't it be up to whatever is being used at the ItemsSource to implement all of the selectioninfo logic? All we need to do is notify of selection changes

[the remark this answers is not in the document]
kazo0: Similar to what we do with selectioninfo in the selector, no? 

https://github.com/unoplatform/uno/blob/master/src%2FUno.UI%2FUI%2FXaml%2FControls%2FPrimitives%2FSelector_Partial.cs#L58-L66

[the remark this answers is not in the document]
kazo0: > Also, I feel like there is something missing for when the VM mutate the selection on their side. How can we be aware of that here? and to update the ui?
> Probably they would ping us via INotifyCollectionChanged somehow or idk? We should check with the MVUX team/their impl.

I discussed with David about the mvux impl here, I was under the impression that we just need to handle selection notification:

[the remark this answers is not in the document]
Xiaoy312: yea, also resynchronizing the selection (vm -> v) after having called De/SelectRanges().
since the vm may reject the selection, the ui must updates accordingly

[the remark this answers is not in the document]
kazo0: Ah, I see, I missed that. So basically what is being done here by the Selector? https://github.com/unoplatform/uno/blob/f912d11baf705517acb7f1c62cb83a47b81b7250/src/Uno.UI/UI/Xaml/Controls/Primitives/Selector_Partial.cs#L116

src/Uno.Toolkit.RuntimeTests/Tests/ItemsRepeaterExtensionsTests.ISelectionInfo.cs
Xiaoy312, line 1:
```
using static some_namespace...ItemsSelectionMode;
```
this hsould help with making the attributes more readable

src/Uno.Toolkit.UI/Behaviors/ItemsRepeaterExtensions.cs
```diff
@@ -454,17 +491,19 @@ private static bool TrySetDefaultSelection(ItemsRepeater ir)
...
 		return false;
 	}
 	private static bool SetEmptySelection(ItemsRepeater ir)
 	{
-		return SetSelectionStates(ir, indexes: null);
+		return SetSelectionStates(ir, null);
 	}
-	private static bool SetSelectionStates(ItemsRepeater ir, params int[]? indexes)
+
+	private static bool SetSelectionStates(ItemsRepeater ir, int? index) => SetSelectionStates(ir, index == null ? null : new[] { index.Value });
```
Xiaoy312, line 505:
```suggestion
	private static bool SetSelectionStates(ItemsRepeater ir, int index) => SetSelectionStates(ir, new[] { index.Value });
```
there is no such thing as null index, only -1

src/Uno.Toolkit.UI/Behaviors/ItemsRepeaterExtensions.cs
```diff
@@ -474,18 +513,54 @@ private static bool SetSelectionStates(ItemsRepeater ir, params int[]? indexes)
...
 		return true;
 	}
 
+	private static void SetSelectionInfoRanges(ItemsRepeater ir, int[] selectedIndexes, int[]? diffIndexes = null)
+	{
+		if (ir.ItemsSource is not ISelectionInfo selectionInfo) return;
+
+		IEnumerable<int>? indexesToDeselect = null;
+
+		// diffIndexes is used to optimize the selection change, by deselecting only the elements that are not in the new selection.
```
Xiaoy312, line 537:
```suggestion
		// diffIndexes is used to optimize the selection change, by updating only the elements that are not in the new selection.
```
not just select, but also deselect

src/Uno.Toolkit.UI/Behaviors/ItemsRepeaterExtensions.cs
```diff
@@ -474,18 +513,54 @@ private static bool SetSelectionStates(ItemsRepeater ir, params int[]? indexes)
...
+			indexesToDeselect = selectionLookup[false];
+			selectedIndexes = selectionLookup[true].ToArray();
+		}
+		else
+		{
+			var selectedRanges = selectionInfo.GetSelectedRanges();
+
+			indexesToDeselect = selectedRanges.SelectMany(ExpandRange)
+					.Where(toDeselect => !selectedIndexes.Contains(toDeselect));
+		}
```
Xiaoy312, line 551: not optimizing the selection range?
selectedRanges vs selectedIndex would be most likely the same for multi-select, with a single element in difference needed to update

src/Uno.Toolkit.UI/Behaviors/ItemsRepeaterExtensions.cs
```diff
@@ -561,4 +637,30 @@ internal static void SetItemSelection(DependencyObject x, bool value)
...
+				n = 1;
+			}
+		}
+
+		if (n > 0)
+		{
+			yield return new(first, n);
+		}
+	}
+	internal static IEnumerable<int> ExpandRange(ItemIndexRange range) => Enumerable.Range(range.FirstIndex, (int)range.Length);
```
Xiaoy312, line 665:
```suggestion
	}

	internal static IEnumerable<int> ExpandRange(ItemIndexRange range) => Enumerable.Range(range.FirstIndex, (int)range.Length);
```

Xiaoy312, line 665: add an overload that takes ItemIndexRange[], so you dont need the SelectMany

Xiaoy312: the return type can also be changed to int[] too

Xiaoy312 (approved):

PR #1071, opened 2024-03-10
fix: wasm splashscreen support

kazo0:


src/Uno.Toolkit.UI/Controls/ExtendedSplashScreen/ExtendedSplashScreen.cs
```diff
@@ -91,7 +91,7 @@ private async Task LoadNativeSplashScreen()
 	}
 
 
-#if !__ANDROID__ && !__IOS__ && !(WINDOWS || WINDOWS_UWP) && !NETSTANDARD2_0
+#if !__ANDROID__ && !__IOS__ && !(WINDOWS || WINDOWS_UWP) && !NET7_0
```
Youssef1313, line 94: I think `NET7_0` will always be true. So, this code is now dead?

src/Uno.Toolkit.UI/Controls/ExtendedSplashScreen/ExtendedSplashScreen.Windows.NetStandard.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if WINDOWS || WINDOWS_UWP || NET472_OR_GREATER || NETSTANDARD2_0
+﻿#if WINDOWS || WINDOWS_UWP || NET472_OR_GREATER || NET7_0
```
Youssef1313, line 1: I'm thinking this should be:

```suggestion
﻿#if WINDOWS || WINDOWS_UWP || UNO_REFERENCE_API
```

src/Uno.Toolkit.UI/Controls/ExtendedSplashScreen/ExtendedSplashScreen.cs
kazo0: Yeah I'm not quite sure about this, but what was happening was that WASM was hitting this method impl when it shouldn't be.

Youssef1313: Yup. Originally Wasm wouldn't hit this because it was `NETSTANDARD2_0`

Thinking about this again, this block of code appears that it should only be true for macOS.

What I would do here is:

- Removing lines 94 to 101 from this file, and move that to a new partial in a new file `ExtendedSplashScreen.macOS.cs`.
- Renaming `ExtendedSplashScreen.Windows.NetStandard.cs` to `ExtendedSplashScreen.Windows.crossruntime.cs`
- Adjust the condition in `ExtendedSplashScreen.Windows.crossruntime.cs` to `﻿#if WINDOWS || WINDOWS_UWP || UNO_REFERENCE_API`
- See if it passes CI.

kazo0: What about for Skia? I believe it should be true for Skia too no?

Youssef1313: UNO_REFERENCE_API should be true on Skia

src/Uno.Toolkit.UI/Controls/ExtendedSplashScreen/ExtendedSplashScreen.Windows.crossruntime.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if WINDOWS || WINDOWS_UWP || NET472_OR_GREATER || NETSTANDARD2_0
```
Youssef1313, line 1: CI failure:

> D:\a\1\s\src\Uno.Toolkit.UI\Controls\ExtendedSplashScreen\ExtendedSplashScreen.Windows.crossruntime.cs(40,15): error CS0102: The type 'ExtendedSplashScreen' already contains a definition for 'SplashIsEnabled' [D:\a\1\s\src\Uno.Toolkit.UI\Uno.Toolkit.UI.csproj::TargetFramework=net7.0-ios]

As it complains about this file, it means the `#if` is true on iOS. However, WINDOWS, WINDOWS_UWP, UNO_REFERENCE_API all should be false on iOS 😨

Youssef1313: Ok you're hitting a bug that was fixed in https://github.com/unoplatform/uno/pull/14282.

To avoid upgrading Uno, I think you can change `UNO_REFERENCE_API` to `__SKIA__ || __WASM__`

Youssef1313: Actually, no, `__SKIA__ || __WASM__` isn't going to work either

Youssef1313: Maybe add `<DefineConstants Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == ''">$(DefineConstants);__SKIA_OR_WASM__</DefineConstants>`

Then use `__SKIA_OR_WASM__` instead of `UNO_REFERENCE_API`. It would also be good to add a comment on that line saying that `UNO_REFERENCE_API` can be used once the project is upgraded to Uno 5.1

kazo0: weird, its compiling fine locally for ios

src/Uno.Toolkit.UI/Controls/ExtendedSplashScreen/ExtendedSplashScreen.macOS.cs
```diff
@@ -0,0 +1,15 @@
+﻿#if !__ANDROID__ && !__IOS__ && !(WINDOWS || WINDOWS_UWP) && !UNO_REFERENCE_API
+using System.Threading.Tasks;
+
+namespace Uno.Toolkit.UI;
+
+public partial class ExtendedSplashScreen
+{
+	private static Task<FrameworkElement?> GetNativeSplashScreen(SplashScreen? splashScreen)
```
lukeblevins, line 8:
```suggestion
	private static Task<FrameworkElement?> GetNativeSplashScreen()
```
Update for post #1061

src/Uno.Toolkit.UI/Controls/ExtendedSplashScreen/ExtendedSplashScreen.Windows.crossruntime.cs
kazo0: > Maybe add `<DefineConstants Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == ''">$(DefineConstants);__SKIA_OR_WASM__</DefineConstants>`
> 
> Then use `__SKIA_OR_WASM__` instead of `UNO_REFERENCE_API`. It would also be good to add a comment on that line saying that `UNO_REFERENCE_API` can be used once the project is upgraded to Uno 5.1

@Youssef1313 Is the condition for that `DefineConstants` supposed to be `"$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == ''"` or should it be `"$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'net7.0'"`?

kazo0: Oh no nevermind I see what you're doing

Youssef1313 (approved):

PR #1130, opened 2024-05-09
fix: selecteditem logic when itemssource is not tabbaritem

kazo0:
Fix for case when you are using an ItemsSource that is not a source of TabBarItem directly. 

Setting the SelectedItem would not work previously


Xiaoy312 (approved):

agneszitte (approved):

agneszitte:
@kazo0 needs to be backported?

kazo0:
@agneszitte not sure, this is something I noticed while using it for an internal project. It's not fixing a regression of anything, just a normal bug

agneszitte:
> @agneszitte not sure, this is something I noticed while using it for an internal project. It's not fixing a regression of anything, just a normal bug

Ok thank you for the feedback. We will backport it as needed then.

PR #1213, opened 2024-08-01
chore: remove problematic uitests

kazo0:


Xiaoy312 (approved):
instead of removal, couldnt we just ignore them with a note to revive later (albeit in different form)

PR #1354, opened 2025-02-23
fix: hot reload ios support for navigationbar

kazo0:
fixes https://github.com/unoplatform/private/issues/617
fixes https://github.com/unoplatform/private/issues/645

Handles Hot Reload scenario where the content of a Page is copied over to a new Page instance and is then set as the Content on the original Page instance.

To avoid an invalid re-render of the UINavigationBar, clear the renderer registrations of the original NavigationBar when unloaded. The NativeFrameHelper will automatically re-wire the renderers when the wrapping UIViewController is about to appear.

### DIFF
--- a/src/Uno.Toolkit.UI/Controls/NavigationBar/Renderer.cs
+++ b/src/Uno.Toolkit.UI/Controls/NavigationBar/Renderer.cs
@@ -51,12 +51,22 @@ namespace Uno.Toolkit.UI
 #if __IOS__
 			if (element is FrameworkElement fe)
 			{
-				fe.Unloaded += (s, e) => Dispose();
+				fe.Unloaded += OnElementUnloaded;
+				_subscriptions.Add(() => fe.Unloaded -= OnElementUnloaded);
+
 			}
 #endif
 
 			_element = new WeakReference<TElement>(element);
 		}
+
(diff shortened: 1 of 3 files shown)


src/Uno.Toolkit.UI/Controls/NavigationBar/Renderer.cs
```diff
@@ -48,6 +48,13 @@ public Renderer(TElement element)
 				throw new ArgumentNullException(nameof(element));
 			}
 
+#if __IOS__
+			if (element is FrameworkElement fe)
+			{
+				fe.Unloaded += (s, e) => Dispose();
+			}
+#endif
```
kazo0, line 56: For this, we needed to ensure that any registered renderers do not attempt to Render() if they are on a Page that is not currently displayed. The iOS nativeframepresenter will re-wire up the renderers when a page is about to appear

But this'd be holding a ref to Element now instead of just the WeakReference. So I don't know if this is a good enough workaround

jeromelaban: You may want to remove the presenter after the dispose has been run to avoid the leak?

jeromelaban (approved):

PR #1489, opened 2025-10-28
docs: add concise InputExtension doc

kazo0:


doc/helpers/walkthroughs/InputExtensions-concise.md
```diff
@@ -0,0 +1,36 @@
...
+```csharp
+<TextBox x:Name="InputA" utu:InputExtensions.AutoFocusNext="True" />
+<TextBox x:Name="InputB" utu:InputExtensions.AutoFocusNext="True" />
+```
+
+AutoFocusNext uses the platform focus manager to find the next element.
+
+#### Explicitly set which input control should receive the focus after pressing Enter
+
+```csharp
```
booskander, line 22: Does it matter if the code is surrounded with csharp or xml formatter ?

PR #1569, opened 2026-04-16
feat: Add GridExtensions (Auto + Direction attached properties)

### LINKED ISSUE #1474: ♥️ Automatic Grid.Row/Column Extension
We should have a behavior that should be able to automatically set the `Grid.Row` and `Grid.Column` properties on an element when being placed inside of a Grid at runtime.

The Grid should have its `RowDefinitions` and/or `ColumnDefinitions` set and we should be able to handle Row/ColumnSpan when setting the Row/Columns on the children

Xiaoy312:
GitHub Issue (If applicable): closes #1474

## PR Type

- Feature

No built-in way to auto-assign `Grid.Row` / `Grid.Column` to children based on their order in the `Children` collection.

## What is the new behavior?

Introduces `GridExtensions` with a single attached property:

- `GridExtensions.Auto` (`bool`) — enables/disables auto-placement

Placement is recalculated whenever children or definitions change (via `LayoutUpdated`, scoped to the `Grid`'s `Loaded`/`Unloaded`).

### Fill Behavior

Fill order is implied by which definitions are present; there is no explicit orientation knob.

| ColumnDefinitions | RowDefinitions | Placement |
|---|---|---|
| Empty | Empty | All children placed at Row 0, Column 0. |
| Defined | Empty | Children fill along columns only (`row=0`, `col = index % cols`). |
| Empty | Defined | Children fill along rows only (`col=0`, `row = index % rows`). |
| Defined | Defined | Two-axis fill, left-to-right wrapping to the next row. Overflow wraps back modulo `rows × cols`. |

### Usage

```xml
<Grid utu:GridExtensions.Auto="True">
    <Grid.ColumnDefinitions>
        <ColumnDefinition /><ColumnDefinition /><ColumnDefinition />
    </Grid.ColumnDefinitions>
    <Grid.RowDefinitions>
        <RowDefinition Height="Auto" /><RowDefinition Height="Auto" />
    </Grid.RowDefinitions>

    <TextBlock Text="(0,0)" />
    <TextBlock Text="(0,1)" />
    <TextBlock Text="(0,2)" />
    <TextBlock Text="(1,0)" />
</Grid>
```

### Limitations

Documented in `doc/helpers/grid-extensions.md` and covered by tests:

- `Grid.RowSpan` / `Grid.ColumnSpan` on children are left untouched but are **not** accounted for when placing — children are positioned as 1×1, so a spanned child will be overlapped by the next one.
- Author-set `Grid.Row` / `Grid.Column` are overwritten while `Auto="True"`; there is no per-child opt-out.
- `GridExtensions` never adds or removes `RowDefinition`/`ColumnDefinition` entries — you define the grid dimensions.

Also includes:
- Sample page (`GridExtensionsSamplePage`) following the `SamplePageLayout` / `ControlExample` pattern
- Runtime tests covering fill order, overflow wrapping, dynamic child add/remove, definition changes, detach/re-attach, layout offsets, and the documented limitations
- Documentation at `doc/helpers/grid-extensions.md`, registered in `doc/toc.yml`

## PR Checklist

- [x] Tested code with current [supported SDKs](../README.md#supported)
  - [ ] UWP
  - [ ] WinUI
  - [ ] iOS
  - [ ] Android
  - [ ] WASM
  - [ ] MacOS
- [x] Updated the documentation as needed:
  - [x] [Extensions Doc Update](https://github.com/unoplatform/uno.toolkit.ui/tree/main/doc/helpers)
- [x] [Runtime Tests and/or UI Tests](https://platform.uno/docs/articles/contributing/guidelines/creating-tests.html) for the changes have been added (for bug fixes / features) (if applicable)
- [x] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

An explicit orientation property (`Direction`, later `AutoOrientation`) was prototyped and then dropped to keep the initial surface to a single `Auto` flag. The implicit semantics above are locked in the docs, so re-introducing an explicit orientation later stays additive.


src/Uno.Toolkit.UI/Behaviors/GridExtensions.cs
```diff
@@ -0,0 +1,153 @@
...
+		if (hash == oldHash) return;
+		SetStateHash(grid, hash);
+
+		if (!GetAuto(grid)) return;
+		
+		// if there is no more than 1 cells, there is nothing to arrange
+		var rowCount = Math.Max(grid.RowDefinitions.Count, 1);
+		var columnCount = Math.Max(grid.ColumnDefinitions.Count, 1);
+		if (rowCount == 1 && columnCount == 1) return;
+
```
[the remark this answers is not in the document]
Xiaoy312: no need

src/Uno.Toolkit.UI/Behaviors/GridExtensions.cs
```diff
@@ -0,0 +1,153 @@
...
+		for (var i = 0; i < childCount; i++)
+		{
+			var (row, column) = fillByColumnsFirst
+				? (i / columnCount % rowCount, i % columnCount)
+				: (i % rowCount, i / rowCount % columnCount);
+
+			Grid.SetRow(grid.Children[i], row);
+			Grid.SetColumn(grid.Children[i], column);
+		}
```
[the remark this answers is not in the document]
Xiaoy312: we assume no row/column-span

doc/helpers/grid-extensions.md
kazo0, line 1: Review Panel — dev/xygu/20260416/auto-grid-ext                                                                                                   
                                                                                                                                                   
  Scope: 2e5c51548bd65b21d2db60f9c0776d6c5984483c..HEAD — adds GridExtensions attached-property family, runtime tests, sample page, and doc.       
                                                                                                                                                   
  Findings                                                                                                                                         
                                                                                                                                                   
  Blocker                                                                                                                                          
                                                                                                                                                   
  1. LayoutUpdated is the wrong change-detection mechanism — architect, security, skeptic                                                          
  - src/Uno.Toolkit.UI/Behaviors/GridExtensions.cs:68-70,97-110                                                                                    
  - Fires on every layout pass for every FrameworkElement in the tree, then writes Grid.Row/Grid.Column from inside the callback — that's exactly
  the re-entrant invalidation pattern this event is notorious for. Per-pass ComputeStateHash walks grid.Children and allocates a HashCode on every
  tick (AGENTS.md §2 hot-path/WASM concern). Security adds: a consumer subclass with an unstable GetHashCode override (e.g. tick-based) produces a
  fresh hash on every pass, re-entering forever — UI-thread DoS, fully blocks WASM. The XML comment justifying this approach claims
  UIElementCollection/DefinitionCollection don't implement IObservableVector, which skeptic flags as likely inaccurate on Uno (and worth
  re-checking on WinAppSDK).
  - Direction: replace with Children.VectorChanged (verify availability first) scoped to Loaded/Unloaded, or reframe as a Panel-derived control
  (AutoGrid : Grid) where OnChildrenChanged/MeasureOverride give structural change tracking for free. Add a re-entrancy guard regardless.

  2. API shape isn't settled — architect, skeptic
  - src/Uno.Toolkit.UI/Behaviors/GridExtensions.cs:15, doc/helpers/grid-extensions.md:21-26
  - Rename history: AutoGrid → GridExtensions, Mode → Auto+Direction → AutoOrientation → hidden → removed. Public surface today is a single Auto
  bool on a plural-named "Extensions" type. Doc's "Fill Behavior" describes implicit orientation driven by which definitions exist, with no
  explicit knob — vertical-major fill will be the first feature request, and re-introducing it later risks silently breaking the implicit
  semantics.
  - Direction: commit to one shape before shipping. Either lock current semantics in docs and ship Auto-only, or restore an explicit Orientation
  (default Horizontal).

  High

  3. No Loaded/Unloaded lifecycle scoping; subscription leaks — security, skeptic
  - src/Uno.Toolkit.UI/Behaviors/GridExtensions.cs:57-81
  - Subscription is acquired on Auto=true and only released on Auto=false. A consumer who removes the Grid from the tree without disabling Auto
  retains the handler closure for the Grid's lifetime — irreversible heap growth on WASM. Also, the closure-stored-as-IDisposable on a DP rooted to
   the Grid extends handler lifetime.
  - Direction: scope subscription to Loaded/Unloaded, or hold it via ConditionalWeakTable<Grid, State> / WeakAttachedDictionary (the codebase
  precedent — see src/Uno.Toolkit.UI/Behaviors/SelectorExtensions.cs:40). Add a LeakTest-style assertion that an Auto-enabled grid is collectable
  post-detach.

  4. Children with user-set Grid.Row/Column/RowSpan/ColumnSpan are silently overwritten — skeptic
  - src/Uno.Toolkit.UI/Behaviors/GridExtensions.cs (UpdateLayout)
  - No detection of pre-set values, no opt-out attached property. Grid.ColumnSpan="2" is preserved but the column position is stomped to 0 and the
  next child overlaps it. Tests don't cover this.
  - Direction: add a GridExtensions.IgnoreAuto opt-out attached property (skip those children) or explicitly document the limitation and add a test
   enshrining the override behavior.

  Medium

  5. Internal state stored as registered attached DPs — architect, security
  - src/Uno.Toolkit.UI/Behaviors/GridExtensions.cs:34-53
  - StateHashProperty and SubscriptionProperty are RegisterAttached'd under owner type GridExtensions with public XAML-resolvable names; a
  third-party theme can write local:GridExtensions.Subscription="{x:Null}" and orphan the live LayoutUpdated handler, or clobber StateHash to force
   re-runs. The codebase already uses WeakAttachedDictionary<TOwner, TKey> (SelectorExtensions.cs:40) for exactly this internal-state pattern.
  - Direction: swap to WeakAttachedDictionary or ConditionalWeakTable<Grid, State>.

  6. Tests assert attached-property values, not layout outcomes — skeptic
  - src/Uno.Toolkit.RuntimeTests/Tests/GridExtensionsTests.cs
  - Every assertion uses Grid.GetRow/Grid.GetColumn. None checks ActualOffset / TransformToVisual. The doc promises "placement logic"; tests don't
  verify placement happened.
  - Direction: at least one E2E test asserting child position after layout.

  7. When_Auto_Enabled_Then_Disabled_Positions_Are_Not_Reset is a weak guard — skeptic
  - src/Uno.Toolkit.RuntimeTests/Tests/GridExtensionsTests.cs
  - Asserts (0,1) for index 1 — exactly the value Auto would have set. Doesn't prove the subscription was torn down.
  - Direction: after disabling, add a child / await idle, assert it stays at default.

  8. Missing coverage: RowSpan/ColumnSpan, pre-set positions, RTL, nested Grids, non-FE children — skeptic
  - src/Uno.Toolkit.RuntimeTests/Tests/GridExtensionsTests.cs
  - The if (... is FrameworkElement) guard silently skips non-FE UIElement children — untested.

  9. Stale/inaccurate comment about IObservableVector / VectorChanged — skeptic
  - src/Uno.Toolkit.UI/Behaviors/GridExtensions.cs:65-67
  - Comment justifies the entire LayoutUpdated approach. Worth verifying on both Uno and WinAppSDK before shipping a comment that locks in the
  design.

  Low

  10. Doc not registered in doc/toc.yml — architect — page is unreachable from nav. Add an entry between flipview-extensions.md and
  input-extensions.md.

  11. No TestPages/ companion / no XAML-driven test — architect — 6c3e736b ("reset StateHash when Auto is disabled") is exactly the kind of
  state-machine bug a XAML-loaded test page would catch. AGENTS.md §5 calls for one when warranted.

  12. ComputeStateHash returns 0 in two meaningful states — skeptic — Auto=false and rows<=1 && cols<=1 both return 0, fragile sentinel.

  13. Sample-page bug: optionAuto.Click reads IsChecked synchronously — skeptic —
  samples/Uno.Toolkit.Samples/Content/Controls/GridExtensionsSamplePage.xaml.cs — should use Checked/Unchecked. Sample-only.

  Info

  - Internal-vs-public test class inconsistency across the suite (GridExtensionsTests is internal, defensible).
  - Default GetHashCode collisions are theoretically possible — see Blocker #1 for the worse variant (custom override).

  Verdict

  block-merge. Architect: needs-redesign. Security: fix-before-merge. Skeptic: fix-first. The two blockers (LayoutUpdated mechanism + unsettled
  - Comment justifies the entire LayoutUpdated approach. Worth verifying on both Uno and WinAppSDK before shipping a comment that locks in the
  design.

  Low

  10. Doc not registered in doc/toc.yml — architect — page is unreachable from nav. Add an entry between flipview-extensions.md and
  input-extensions.md.

  11. No TestPages/ companion / no XAML-driven test — architect — 6c3e736b ("reset StateHash when Auto is disabled") is exactly the kind of
  state-machine bug a XAML-loaded test page would catch. AGENTS.md §5 calls for one when warranted.

  12. ComputeStateHash returns 0 in two meaningful states — skeptic — Auto=false and rows<=1 && cols<=1 both return 0, fragile sentinel.

  13. Sample-page bug: optionAuto.Click reads IsChecked synchronously — skeptic —
  samples/Uno.Toolkit.Samples/Content/Controls/GridExtensionsSamplePage.xaml.cs — should use Checked/Unchecked. Sample-only.

  Info

  - Internal-vs-public test class inconsistency across the suite (GridExtensionsTests is internal, defensible).
  - Default GetHashCode collisions are theoretically possible — see Blocker #1 for the worse variant (custom override).

  Verdict

  block-merge. Architect: needs-redesign. Security: fix-before-merge. Skeptic: fix-first. The two blockers (LayoutUpdated mechanism + unsettled
  API) recur across all three lenses; the High findings (#3, #4) are the lifecycle and intent-preservation bugs that LayoutUpdated-based detection
  masks.

kazo0 (approved):

doc/helpers/grid-extensions.md
nickrandolph: @Xiaoy312 are you able to work through the findings here and get this PR mergable

Xiaoy312:
@nickrandolph @kazo0 worked through the panel — here's the triage. Most of it was actionable and is now fixed; a few items I'm closing with reasons rather than code.

### Fixed

| # | Finding | What changed |
|---|---|---|
| B1 | `LayoutUpdated` re-entrancy / per-pass cost | Added a re-entrancy guard around the placement loop. Children are now hashed with `RuntimeHelpers.GetHashCode` (identity hash), so a consumer overriding `GetHashCode` with an unstable implementation can no longer force a placement pass on every layout — that was the DoS variant. Also replaced the interface-typed `foreach` with an indexed loop, so we no longer box an enumerator per layout pass. |
| H3 | No lifecycle scoping; subscription leak | The `LayoutUpdated` subscription is now scoped to the `Grid`'s `Loaded`/`Unloaded`, and the stored hash resets on `Unloaded` so a reload always recomputes. |
| M5 | Internal state as XAML-resolvable attached DPs | `StateHashProperty`/`SubscriptionProperty` are gone. State now lives in a `WeakAttachedDictionary<Grid, string>`, following the `SelectorExtensions` precedent. No more `GridExtensions.Subscription="{x:Null}"` surface. |
| M6 | Tests assert DP values, not layout | Added `When_Auto_Children_Are_Laid_Out_In_Their_Cells`, asserting actual child offsets via `TransformToVisual` on a fixed-size grid. |
| M7 | Weak disable guard | After disabling `Auto`, the test now adds a child and asserts it stays at `(0,0)` — proving teardown rather than just that old values persisted. |
| M8 | Missing coverage | Added tests for definition changes at runtime, detach/re-attach, pre-set positions, and spans. |
| L10 | Doc not in `doc/toc.yml` | Registered between `FlipView-extensions.md` and `input-extensions.md`. |
| L12 | `0` sentinel in `ComputeStateHash` | Early-outs hoisted into `UpdateLayout`; the hash function now only hashes, and the duplicated single-cell check is gone. |
| L13 | Sample page `Click` handler | Switched to `Checked`/`Unchecked`. (For what it's worth, `ToggleButton` updates `IsChecked` before raising `Click`, so this wasn't actually broken — but the event pair is the right one.) |

### Closing without changes

**L9 — the `IObservableVector` comment is accurate, so `LayoutUpdated` stays.** I checked the metadata rather than the comment:

- `RowDefinitionCollection` / `ColumnDefinitionCollection` implement only `IList<T>` / `IEnumerable<T>` — no change notification at all, on Uno *or* in the generated WinUI API surface (`src/Uno.UI/Generated/3.0.0.0/Microsoft.UI.Xaml.Controls/RowDefinitionCollection.cs`).
- `UIElementCollection` implements `INotifyCollectionChanged` on Uno only, and nothing on WinAppSDK — no `IObservableVector`, so no `VectorChanged`.

So `Children.VectorChanged` doesn't exist and definition changes are unobservable on both targets. There's no alternative hook to move to; the concrete concerns behind B1 are addressed above instead.

**H4 / spans — documented rather than implemented.** Placement stays index-based. Spans are left untouched on the child but not accounted for when positioning, and author-set `Grid.Row`/`Column` are overwritten while `Auto="True"`. Both are now called out in a Limitations section in the doc and pinned by `When_Child_Has_Span_It_Is_Ignored` / `When_Child_Has_Preset_Position_It_Is_Overwritten`, so changing the behaviour later breaks a test rather than silently shifting semantics.

**B2 — shipping `Auto`-only.** The explicit orientation property was dropped deliberately. The implicit fill semantics are locked in the doc, so adding an explicit orientation later is additive. Also updated the PR description, which was still documenting the removed `Direction` property.

**L11 — no `TestPages/` companion.** The existing pages there are all navigation-scoped (NavBar / LoadingView / AncestorBinding); a XAML page wouldn't add coverage over the runtime tests here.

One note: the panel references `AGENTS.md §2` and `§5`, but there's no `AGENTS.md` in this repo, so I couldn't check what those two points were pointing at.

Locally on desktop: 14/14 `GridExtensionsTests` pass, and the full suite is 319/323 with the only 4 failures being `DrawerFlyoutTests.OpenDirection_Layout` — which reproduce identically with `GridExtensions.cs` reverted to its pre-fix state, so they're pre-existing here and unrelated (they're green on CI).

doc/helpers/grid-extensions.md
Xiaoy312: Worked through all 13 findings in `27501c94` — full triage in https://github.com/unoplatform/uno.toolkit.ui/pull/1569#issuecomment-5196744185.

Short version: 9 fixed (re-entrancy guard, `Loaded`/`Unloaded` scoping, state moved off XAML-resolvable attached properties, identity hashing, allocation-free loop, sentinel cleanup, `toc.yml`, test coverage, sample page events). 4 closed with reasons — notably L9, where I verified the `IObservableVector` comment is accurate, so there is no `VectorChanged` hook to move to on either Uno or WinAppSDK.

Resolving this thread; happy to reopen if any individual finding needs more.

PR #1591, opened 2026-05-07
feat: replace hardcoded dimensions with design tokens in toolkit controls

### LINKED ISSUE #1592: Replace hardcoded dimensions with design tokens in toolkit controls
Toolkit control styles (Material v2 and Simple) use hardcoded pixel values for padding, margins, corner radii, and heights. These should reference the shared design token system (`Space*`, `Radius*`, `ControlHeight*`, `IconSize*`) so they automatically respond to density and corner-radius configuration changes.

## Scope

- Card, Chip, ChipGroup, Divider, NavigationBar, TabBar (both Material v2 and Simple variants)
- Add design tokens documentation page
- Add regression tests verifying token wiring

Kunal22shah:
Replace hardcoded pixel values (padding, margins, corner radii, heights) with design token references (`Space*`, `Radius*`, `ControlHeight*`, `IconSize*`) across all Material v2 and Simple toolkit control styles. Add a design tokens documentation page and regression tests.

Closes https://github.com/unoplatform/uno.toolkit.ui/issues/1592

### DIFF
--- a/src/Uno.Toolkit.RuntimeTests/Tests/Given_DesignTokens.cs
+++ b/src/Uno.Toolkit.RuntimeTests/Tests/Given_DesignTokens.cs
@@ -19,32 +19,25 @@ namespace Uno.Toolkit.RuntimeTests.Tests;
 [TestClass]
 public class Given_DesignTokens
 {
-	// ─────────────────────────────────────────────────────────────────────
-	// Helpers
-	// ─────────────────────────────────────────────────────────────────────
+	#region Default Density
 
-	private static (Grid container, MaterialToolkitTheme theme) CreateThemedContainer()
+	[TestMethod]
+	[RunsOnUIThread]
+	public void When_DefaultDensity_Then_SpacingBaseIsFour()
 	{
-		var theme = new MaterialToolkitTheme();
-		var container = new Grid();
-		container.Resources.MergedDictionaries.Add(theme);
-		return (container, theme);
+		var (container, theme) = CreateThemedContainer();
+
+		// Default density should be Regular (int value 4), giving spacing base=4.
+		// Validate via Space100 which equals base×1.
+		Assert.AreEqual((int)theme.DefaultDensity, 4,
+			"Default density should be Regular (4)");
+		Assert.AreEqual(4.0, GetResource<double>(container, "Space100"), 0.001,
+			"Space100 should be 4 (base=4 × 1), confirming spacing base=4");
 	}
 
-	private static T GetResource<T>(Grid container, string key)
-	{
-		if (container.Resources.TryGetValue(key, out var value) && value is T typed)
-		{
-			return typed;
-		}
+	#endregion
 
-		Assert.Fail($"Resource '{key}' not found or not of type {typeof(T).Name}");
-		return default!;
-	}
-
-	// ═══════════════════════════════════════════════════════════════════════
-	// 1. CARD — padding and corner radius reference design tokens
-	// ═══════════════════════════════════════════════════════════════════════
+	#region Card
 
 	[TestMethod]
 	[RunsOnUIThread]
@@ -79,9 +72,9 @@ public class Given_DesignTokens
 			"CardElevationMargin should be Space150Thickness (6) at Regular density");
 	}
 
-	// ═══════════════════════════════════════════════════════════════════════
-	// 2. CHIP — height, corner radius, padding, margins reference tokens
-	// ═══════════════════════════════════════════════════════════════════════
+	#endregion
+
+	#region Chip
 
 	[TestMethod]
 	[RunsOnUIThread]
@@ -147,9 +140,9 @@ public class Given_DesignTokens
 			"ChipContentMinHeight should be Space500 (20) at Regular density");
 	}
 
-	// ═══════════════════════════════════════════════════════════════════════
-	// 3. NAVIGATION BAR — heights and icon sizes reference tokens
-	// ═══════════════════════════════════════════════════════════════════════
+	#endregion
+
+	#region NavigationBar
 
 	[TestMethod]
 	[RunsOnUIThread]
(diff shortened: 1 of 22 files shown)


src/library/Uno.Toolkit.Material/Styles/Controls/v2/NavigationBar.xaml
```diff
@@ -68,19 +68,19 @@
 							 Color="Transparent" />
 
 			<x:Double x:Key="MaterialNavigationBarElevation">4</x:Double>
-			<x:Double x:Key="MaterialXamlNavigationBarHeight">64</x:Double>
-			<x:Double x:Key="MaterialNavigationBarHeight">64</x:Double>
-			<Thickness x:Key="MaterialNavigationBarContentMargin">16,0</Thickness>
+			<StaticResource x:Key="MaterialXamlNavigationBarHeight" ResourceKey="Space1600" />
+			<StaticResource x:Key="MaterialNavigationBarHeight" ResourceKey="Space1600" />
+			<StaticResource x:Key="MaterialNavigationBarContentMargin" ResourceKey="Space400HorizontalThickness" />
```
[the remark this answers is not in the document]
Kunal22shah: This is intentional. The NavigationBar height using `Space1600` is designed to scale with density, matching the behavior in Uno.Themes where NavigationBar also uses density-dependent spacing. In Compact mode the nav bar is tighter (48px), Regular is 64px, and Comfy is 80px — providing a consistent density experience throughout the app.

The "fixed" tokens (`ControlHeightSmall`, `ControlHeightLarge`, etc.) are reserved for cases where the control's touch target or minimum interaction area must remain constant. Navigation bar height is a layout/spacing concern that should adapt to density.

src/Uno.Toolkit.RuntimeTests/Tests/Given_DesignTokens.cs
```diff
@@ -0,0 +1,402 @@
...
+
+/// <summary>
+/// Verifies that toolkit control resources reference the correct design tokens
+/// via <see cref="MaterialToolkitTheme"/>.
+/// Note: Toolkit XAML resources using StaticResource aliases resolve at parse time
+/// against app-level resources, so these tests verify Regular density (base=4, corner=4)
+/// which matches the app theme configuration.
+/// </summary>
+[TestClass]
+public class Given_DesignTokens
```
Xiaoy312, line 20: consider grouping the methods with `#region` or `partial class`(same file)
and move helpers to the bottom

Xiaoy312: it seems that you are throwing multiple controls here
consider splitting into multiple `partial class` files: Given_DesignTokens.${control}.cs

edit: but only when there is enough tests to justify it,
like in Given_SimpleDesignTokens.cs where you only have 1-2 tests per controls, same file partial is good enough

src/Uno.Toolkit.RuntimeTests/Tests/Given_DesignTokens.cs
```diff
@@ -0,0 +1,402 @@
...
+	// 1. CARD — padding and corner radius reference design tokens
+	// ═══════════════════════════════════════════════════════════════════════
+
+	[TestMethod]
+	[RunsOnUIThread]
+	public void When_CardPadding_Then_MatchesSpace400Thickness()
+	{
+		var (container, _) = CreateThemedContainer();
+		var padding = GetResource<Thickness>(container, "CardPadding");
+		// Space400Thickness at Regular density (base=4): 4×4=16
```
Xiaoy312, line 48: we are making assuming about the base=4 for pretty much all the test,
should add one more test just to validate that

src/Uno.Toolkit.RuntimeTests/Tests/Given_SimpleDesignTokens.cs
```diff
@@ -0,0 +1,190 @@
...
+	// ─────────────────────────────────────────────────────────────────────
+
+	private static readonly Type? s_simpleThemeType =
+		Type.GetType("Uno.Toolkit.UI.Simple.SimpleToolkitTheme, Uno.Toolkit.WinUI.Simple");
+
+	private static Grid CreateThemedContainer()
+	{
+		if (s_simpleThemeType == null)
+		{
+			Assert.Inconclusive("SimpleToolkitTheme not available — run in Simple sample app");
```
Xiaoy312, line 32: instead of inconclusive, we should either:
- dynamically filter the test by adding some kind of attribute: [TestMethod]->[SimpleThemeTestMethod]
- exclude the cs for non-simple build, which will be hard with current setup

Xiaoy312 (approved):

src/Uno.Toolkit.RuntimeTests/Tests/Given_SimpleDesignTokens.cs
Xiaoy312: this can be its own issue, dont necessarily need to address them here

src/Uno.Toolkit.RuntimeTests/Tests/Given_DesignTokens.cs
Kunal22shah: Done — reorganized `Given_DesignTokens.cs` with `#region` blocks grouped by control (Card, Chip, NavigationBar, TabBar, Divider, Density Invariance, Integration) and moved helpers to the bottom. Kept it in a single file since the test count per control is manageable.

src/Uno.Toolkit.RuntimeTests/Tests/Given_DesignTokens.cs
Kunal22shah: Good call — added `When_DefaultDensity_Then_SpacingBaseIsFour` test that explicitly validates `DefaultDensity` is Regular (4) and `Space100` resolves to 4.0, confirming the base=4 assumption all other tests rely on.

PR #1595, opened 2026-05-11
docs: fix incorrect property names and missing namespace prefix in SafeArea.md

agneszitte:
## What

Fixes two documentation errors in `doc/controls/SafeArea.md`:

1. **Wrong property name in example text**: `InsetMask="Left, Right"` → `Insets="Left, Right"` — the property was renamed from `InsetMask` to `Insets` but this inline example was missed.
2. **Missing `utu:` namespace prefix**: `<SafeArea>` → `<utu:SafeArea>` in the "as a control" XAML snippet — inconsistent with every other usage in the doc.

## Why

Developers copying these snippets would hit XAML compilation errors. The rest of the document (20+ XAML snippets) correctly uses `Insets=` and `utu:SafeArea`, so these were oversights from the original rename.

## Verification

Cross-referenced every property name, enum value, XAML snippet, and behavioral claim in `SafeArea.md` and `SafeArea.howto.md` against the source code (`SafeArea.cs`, `SafeArea.Properties.cs`, `SafeArea.xaml`, `SafeAreaPresenter.cs`, Material `TabBar.xaml` styles). No other discrepancies found.

Note: No related issue (Internal maintenance / documentation accuracy fix).

### DIFF
--- a/build/workflow/stage-determine-changes.yml
+++ b/build/workflow/stage-determine-changes.yml
@@ -37,6 +37,7 @@ jobs:
       foreach ($file in $changedFiles -split "`n") {
         # Identifying changes as documentation:
           $isDoc = $file.StartsWith("doc/") -or               # Files in the 'doc/' directory
+            $file.StartsWith("specs/") -or                    # Files in the 'specs/' directory
             ($file -match "^[^/]+\.md$") -or                  # Markdown files at the root level (with no subdirectories involved)
             ($file -match "^\.github/.*\.md$") -or            # Markdown files within the .github folder (including its subdirectories)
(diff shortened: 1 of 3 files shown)


Xiaoy312 (approved):

ajpinedam (approved):

agneszitte:
Also fixed markdown linting error (`MD040/fenced-code-language`) in `specs/safearea-bounds-guard/recap.md` — added `text` language specifier to the fenced code block.

agneszitte:
Also added `specs/` to the doc-only path detection in `stage-determine-changes.yml` so that spec-only changes don't trigger full CI builds (same treatment as `doc/` and root `.md` files).

PR #1618, opened 2026-07-30
feat(tabbar): add built-in accessibility support

morning4coffe-dev:
GitHub Issue (If applicable): unoplatform/kahua-private#507

## PR Type

- Bugfix

`TabBar` and `TabBarItem` never override `OnCreateAutomationPeer()`. WinUI only creates a peer for a plain `FrameworkElement` when `AutomationProperties.Name` or `LabeledBy` is explicitly set, and even then the result is a generic peer reporting `AutomationControlType.Group`.

The practical consequences are:

- The tab strip is not identified as a tab control to assistive technology.
- Tab items are not identified as tab items, and the selected item exposes no selection state.
- Setting an automation name on a `TabBarItem` appears to do nothing, which is what the linked issue reports.

## What is the new behavior?

Two dedicated automation peers are added under `Controls/TabBar/`, and both controls now override `OnCreateAutomationPeer()`:

- `TabBarAutomationPeer` reports `AutomationControlType.Tab` and implements `ISelectionProvider`. `CanSelectMultiple` is `false` and `IsSelectionRequired` is `false`, matching `SelectedIndex` defaulting to `-1`.
- `TabBarItemAutomationPeer` reports `AutomationControlType.TabItem` and implements `ISelectionItemProvider`. `IsSelected` reflects `TabBarItem.IsSelected`, `Select()` selects the item, `AddToSelection()` is rejected when it would create a multi-selection, and `RemoveFromSelection()` clears the selection when permitted.
- Selection changes raise `SelectionItemPatternIdentifiers.IsSelectedProperty` on the item peer.
- `GetNameCore()` honours an explicit `AutomationProperties.Name` or `LabeledBy`, and otherwise resolves a name from the rendered content template rather than falling back to `ToString()` on the data object.

`TabBarAutomationPeer` intentionally derives from `FrameworkElementAutomationPeer` rather than `ItemsControlAutomationPeer`. `ItemsControlAutomationPeer` creates its own `ItemAutomationPeer` instances for items, which would compete with the dedicated `TabBarItemAutomationPeer` returned by `TabBarItem.OnCreateAutomationPeer()`. Deriving from `FrameworkElementAutomationPeer` keeps a single, stable identity per item.

## Verification

Built on macOS with the single-platform desktop override:

- `dotnet build src/Uno.Toolkit.UI/Uno.Toolkit.WinUI.csproj -c Debug -p:TargetFrameworkOverride=desktop` — 0 warnings, 0 errors
- `dotnet build src/Uno.Toolkit.UI/Uno.Toolkit.WinUI.csproj -c Release -p:TargetFrameworkOverride=desktop` — 0 warnings, 0 errors (Release treats warnings as errors)
- `dotnet build src/Uno.Toolkit.RuntimeTests/Uno.Toolkit.RuntimeTests.csproj -c Debug -p:TargetFrameworkOverride=desktop` — 0 warnings, 0 errors

Runtime tests were added to `TabBarTests.cs` covering the control types, the selection pattern before and after selection, selection toggling across items, rejection of invalid multi-selection, and both the explicit-name and templated-content naming paths.

## Not verified

- Announcement text and event routing with a real screen reader (Narrator, VoiceOver, TalkBack) have not been checked.
- Only the desktop target was built and exercised; Windows, iOS, Android and WASM heads were not.

## PR Checklist

- [x] Tested code with current supported SDKs
	- [ ] UWP
	- [ ] WinUI
	- [ ] iOS
	- [ ] Android
	- [ ] WASM
	- [ ] MacOS
- [ ] Updated the documentation as needed
- [x] Runtime Tests and/or UI Tests for the changes have been added

### DIFF
--- a/src/Uno.Toolkit.UI/Controls/TabBar/TabBar.cs
+++ b/src/Uno.Toolkit.UI/Controls/TabBar/TabBar.cs
@@ -11,6 +11,7 @@ using Windows.Foundation.Collections;
 
 #if IS_WINUI
 using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Automation.Peers;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Controls.Primitives;
 using Microsoft.UI.Xaml.Data;
@@ -19,6 +20,7 @@ using Microsoft.UI.Xaml.Media;
 using Microsoft.UI.Xaml.Navigation;
 #else
 using Windows.UI.Xaml;
+using Windows.UI.Xaml.Automation.Peers;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Controls.Primitives;
 using Windows.UI.Xaml.Data;
@@ -58,6 +60,9 @@ namespace Uno.Toolkit.UI
 			UpdateOrientation();
 			UpdateIndicatorPlacement();
 		}
+
+		/// <inheritdoc />
+		protected override AutomationPeer OnCreateAutomationPeer() => new TabBarAutomationPeer(this);
 
 		protected override bool IsItemItsOwnContainerOverride(object? item) => item is TabBarItem;
 
@@ -447,6 +452,31 @@ namespace Uno.Toolkit.UI
 			}
 		}
 
+		internal bool TryClearSelection(TabBarItem item)
+		{
+			if (!ReferenceEquals(item, GetSelectedTabBarItem()))
+			{
+				return false;
+			}
+
+			var previouslySelectedItem = SelectedItem;
+
+			try
+			{
+				_isSynchronizingSelection = true;
+				item.IsSelected = false;
+				SetValue(SelectedItemProperty, null);
+				SelectedIndex = -1;
+			}
+			finally
+			{
+				_isSynchronizingSelection = false;
+			}
+
+			RaiseSelectionChangedEvent(previouslySelectedItem, null);
+			return true;
+		}
+
 		private void RaiseSelectionChangedEvent(object? prevItem, object? nextItem)
 		{
 			var eventArgs = new TabBarSelectionChangedEventArgs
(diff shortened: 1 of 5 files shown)


src/Uno.Toolkit.UI/Controls/TabBar/TabBarItemAutomationPeer.cs
```diff
@@ -0,0 +1,151 @@
...
+					var name = GetContentName(child);
+					if (!string.IsNullOrEmpty(name))
+					{
+						return name;
+					}
+				}
+			}
+
+			return string.Empty;
+		}
```
Xiaoy312, line 151:
```suggestion
	private static string GetContentName(UIElement element)
	{
		return GetContentNameCore(element) is { Length: > 0 } name ? name : null
			?? element.EnumerateDescendants()// depth-first
				.OfType<UIElement>()
				.Select(GetContentNameCore)
				.FirstOrDefault(x => !string.IsNullOrEmpty(x))
			?? string.Empty;
	
	
	private static string GetContentNameCore(UIElement element)
	{
		if (FrameworkElementAutomationPeer.CreatePeerForElement(element) is { } peer)
		{
			var name = peer.GetName();
			if (!string.IsNullOrEmpty(name))
			{
				return name;
			}
		}
		
		return string.Empty;
	}
```

src/Uno.Toolkit.UI/Controls/TabBar/TabBarItemAutomationPeer.cs
```diff
@@ -0,0 +1,151 @@
...
+				throw new InvalidOperationException("The TabBarItem is not selectable.");
+			}
+
+			tbi.IsSelected = true;
+		}
+
+		/// <inheritdoc />
+		public void AddToSelection()
+		{
+			if (Owner is not TabBarItem tbi || tbi.IsSelected)
```
Xiaoy312:
```suggestion
			if (Owner is not TabBarItem { IsSelected: false })
```
keep it consistent with RemoveFromSelection?

Xiaoy312 (approved):

src/Uno.Toolkit.UI/Controls/TabBar/TabBar.cs
```diff
@@ -471,6 +501,27 @@ private void RaiseSelectionChangedEvent(object? prevItem, object? nextItem)
...
+				return selectedItem;
+			}
+
+			foreach (var container in this.GetItemContainers<UIElement>())
+			{
+				if (GetInnerContainer(container) is { IsSelected: true } selectedContainer)
+				{
+					return selectedContainer;
+				}
+			}
```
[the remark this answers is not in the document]
morning4coffe-dev: Keeping the explicit loop here. This selection lookup is allocation-free and returns immediately on the first selected container; a LINQ chain would add enumeration and delegate overhead without changing behavior.

morning4coffe-dev:
recheck

Issue #1300, opened 2024-12-06
[Canaries] Failing Android ChipGroup Selection Tests


kazo0:
Issue to track test failures on the canaries for the ChipGroup Selection Tests

> 6 unit test(s) failed (count=6).
> 	Failing Tests:
> 	1. VariousMode_TapSelection (Single,System.Int32[],0)
> 	2. VariousMode_TapSelection (Single,System.Int32[],1)
> 	3. VariousMode_TapSelection (Single,System.Int32[],1)
> 	4. VariousMode_TapSelection (Single,System.Int32[],2)
> 	5. Single_SetSelection()
> 	6. Initial_Selection()
> 
> 
> ---
> 	Details:
> Failed: VariousMode_TapSelection (Single,System.Int32[],0) [AssertFailedException] 
>  Assert.AreEqual failed. Expected:<0>. Actual:<(null)>. 
> 
> Failed: VariousMode_TapSelection (Single,System.Int32[],1) [AssertFailedException] 
>  Assert.AreEqual failed. Expected:<0>. Actual:<(null)>. 
> 
> Failed: VariousMode_TapSelection (Single,System.Int32[],1) [AssertFailedException] 
>  Assert.AreEqual failed. Expected:<0>. Actual:<(null)>. 
> 
> Failed: VariousMode_TapSelection (Single,System.Int32[],2) [AssertFailedException] 
>  Assert.AreEqual failed. Expected:<0>. Actual:<(null)>. 
> 
> Failed: Single_SetSelection() [AssertFailedException] 
>  Assert.AreEqual failed. Expected:<0>. Actual:<(null)>. 
> 
> Failed: Initial_Selection() [AssertFailedException] 
>  Assert.AreEqual failed. Expected:<Uno.Toolkit.UI.Chip>. Actual:<(null)>. SelectedItem is expected to be container#0

PR #1635, opened 2026-09-03
chore!: Uno Platform 7.0 retarget (SDK, net10.0, Skia mobile port)

MartinZikmund:
**GitHub Issue:** relates to unoplatform/uno-private#2074

## PR Type

- Build or CI related changes
- Bugfix
- Refactoring

The toolkit targets Uno Platform 6 (`net9.0`) and does not build or run against Uno Platform 7.0.

## What is the new behavior?

Uno.Toolkit builds and runs on **Uno Platform 7.0** (`Uno.Sdk.Private` `7.0.0-dev.701`, `net10.0`), on top of the Uno 7 builds of Uno.Themes (`9.0.0-dev.15`) and C# Markup (`7.0.0-dev.33`).

The full plan, the decisions behind it and what is still open are in `specs/uno7-groundwork/progress.md`.

### Target frameworks and the Skia mobile heads

- Libraries keep `net10.0`, `net10.0-android`, `net10.0-ios` and `net10.0-windows10.0.19041`, matching Uno.Themes 9.0. Mac Catalyst is gone.
- In Uno 6, a Skia Android/iOS head swapped every library's mobile asset for its plain `netX.0` build, so Skia users have always run the plain toolkit build. Uno 7 (`unoplatform/uno@bc6254848a`) stops that swap. The mobile assets are therefore rebuilt to **behave like the plain Skia build**, and the native-renderer code they used to carry is removed.

### Native rendering removal (breaking)

- `NavigationBar` uses its XAML template on every platform. `NativeFramePresenter`, `NativeNavigationBarPresenter`, the native renderers, `NativeDefaultToolkitFrame`, the `*NativeNavigationBarTemplate` keys and the `IsNativeStyle` frame styles are removed. `doc/controls/NavigationBar.md` has a migration section.
- The TabBar, FlipView, Chip and segmented-control styles collapse their `mobile:` forks onto the arm Skia heads already used.
- `TabBarSelectorBehavior` follows the FlipView's `ScrollViewer` everywhere. `ScrollableHelper` gets a managed scroll-to-top/bottom, where the mobile path previously threw `NotImplementedException`.
- `ExtendedSplashScreen` uses the Skia splash on Android and iOS, built from `UnoSplash.def`. `Init(Activity)` keeps its signature and still hands off to the AndroidX splash screen; this is the one native-only behavior kept, as an OS-level service.
- `SafeArea` computes insets from `Bounds` and `VisibleBounds` on every Uno target. The removed `IsStatusBarTranslucent` branch goes, and the Android bounds-race guard stays.

### Other Uno 7 fallout

- BC53 (`Uno.UI.Toolkit` → `Uno.UI.Extras` namespaces) and BC26 (`DependencyObject` is a class).
- `IDependencyObjectStoreProvider` → `MarkupHelper.SetParent`.
- `ShadowContainer` renders through `SKCanvasElement` on Uno targets. `SkiaSharp.Views.Uno.WinUI` is still built against Uno 5 and stopped the app from starting, so the reference is gone; WinAppSDK keeps `SKXamlCanvas`.
- `DrawerControl.IsOpen` set off the UI thread is marshalled to it. Uno 7 now enforces thread access in composition, as WinUI does.
- Sample heads: the Android host goes through `UnoPlatformHostBuilder`, and Hot Design / App MCP are disabled in Debug until they ship Uno 7 builds.
- CI: the .NET SDK is pinned to `10.0.101`, the band `uno.check` provisions `wasm-tools` for (`NETSDK1147`, same fix as Uno.Themes), and the install cache is now keyed on that version.
- Hot reload: the dev-server loads every library flavor and ignores the build's global properties, so libraries evaluate `net10.0` only when `UnoIsHotReloadHost` is set, and the CI job writes `crosstargeting_override.props` for desktop. The Debug heads call `EnableHotReload()`, and `HotReloadTestHelper` waits for the edited assembly's delta, since Uno 7 also emits a delta for the head's `uno.hot-reload.info`.
- iOS trim analyzer: `IL2122` is fixed by assembly-qualifying the name; `IL2070`/`IL2075` are suppressed with justification on name-based DP lookups.

## Verification

**CI:** every job on this PR is green, including Packages, all sample builds, and the desktop and hot-reload runtime tests.

Run locally on Windows with the commands the pipeline uses:

| Leg | Result |
|---|---|
| `Uno.Toolkit-packages.slnf` + `Uno.Toolkit.RuntimeTests`, `net10.0-android` / `net10.0-ios` | 0 errors, no `UNOB0020` |
| Desktop publish: Material, Cupertino, Simple | 0 errors |
| WASM publish: Material, Cupertino | 0 errors |
| Android build: Material, Cupertino | 0 errors |
| Packages (MSBuild, `GeneratePackageOnBuild`) | 0 errors, 8 packages |
| Desktop runtime tests (Linux, WSLg) | 354 of 360 before the last fix; the 6 incremental-loading failures pass after it (re-run with `Attempts: 1`) |
| Hot-reload runtime tests (Debug desktop) | 48 of 48 after the hot-reload fixes (CI had 39 of 45 failing) |
| Skia Android runtime tests (API 34 emulator, Release, local harness) | 244 of 365; all 121 failures are exact-size assertions broken by rounding at 2.625 density. Every port-sensitive test passes: splash, SafeArea system bars, NavigationBar, ChipGroup #1300, drawer, scroll-to-top |

Published assemblies were checked with a metadata reader: nothing references the removed Uno 6 assemblies.

**Not verified here:** Skia iOS runtime tests, the visual splash handoff on devices, the WinAppSDK sample heads, and the iOS app heads, which need a Mac.

## PR Checklist

- [x] Tested code with current supported SDKs
	- [x] WinUI (Skia desktop)
	- [ ] iOS
	- [ ] Android
	- [x] WASM (build)
- [x] Updated the documentation as needed (NavigationBar, ExtendedSplashScreen, SafeArea, AGENTS.md)
- [x] Runtime Tests added or ported (NavigationBar, ScrollableHelper, TabBarItemExtensions, ExtendedSplashScreen, SafeArea)
- [ ] Contains **NO** breaking changes — breaking changes are described in the `BREAKING CHANGE:` bodies of the `!` commits

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SQBL3dpwkP1vsVzy7Kcijc


samples/Uno.Toolkit.Samples.Cupertino/CupertinoSampleApp.csproj
```diff
@@ -2,7 +2,7 @@
 	<!--
 		Expand platform suffix overrides to versioned TFMs for sample app (net10.0).
 		This allows crosstargeting_override.props to use platform suffixes (e.g., 'ios', 'android')
```
agneszitte, line 4: @MartinZikmund do not merge this PR yet, please

I am currently validating the versions and servicing branch with @kazo0, and the major bump + servicing branch may change for this repo. 

I will resolve this comment as soon as it is validated with him, and I will give you the updated details ;)

agneszitte (dismissed):

samples/Directory.Packages.props
```diff
@@ -1,7 +1,6 @@
-<Project ToolsVersion="15.0">
+﻿<Project ToolsVersion="15.0">
```
agneszitte, line 1: Heads-up @MartinZikmund (Cc @kazo0)
`main` is now `11.0-dev.{height}`, not `10.0-dev` (#1637). 10.0 moved to the Uno 6.x line as `servicing/10.0`, because aligning to Uno.Themes 8.0 crosses a major on `UnoThemesVersion`. `servicing/9.2` is retired. This PR doesn't touch `version.json`, so no conflict with the bump.

Please rebase *after* #1636 merges, not before. It's landing on `main` first, and it edits `UnoThemesVersion` inside your `samples/Directory.Packages.props` hunk, so rebasing now means resolving that twice. You'll also inherit the `SimpleFontFamily` → `DefaultFontFamily` swap, which you'd otherwise have to redo here anyway.

Then the last piece is `UnoThemesVersion` → `9.0.0-dev.*`, which needs unoplatform/Uno.Themes#1722 to publish, that's this PR's real blocker.

Nit: the diff adds a UTF-8 BOM to `samples/Directory.Packages.props`, looks unintentional.

src/library/Uno.Toolkit.Material/Styles/Controls/v1/BottomTabBar.xaml
```diff
@@ -3,11 +3,9 @@
 					xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
 					xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
 					xmlns:um="using:Uno.Material"
-					xmlns:toolkit="using:Uno.UI.Toolkit"
-                    xmlns:not_mobile="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-                    xmlns:mobile="http://uno.ui/mobile" 
+					xmlns:toolkit="using:Uno.UI.Xaml.Controls"
```
Xiaoy312, line 6: lets not use something as confusing as "toolkit" in the toolkit repo itself, while refering to something completely unrelated
```suggestion
					xmlns:uuxc="using:Uno.UI.Xaml.Controls"
```

src/library/Uno.Toolkit.Material/Styles/Controls/v1/_Common.xaml
Xiaoy312, line 1: i cant really remember the historical luggage with v1
i believe it is completely unused now<sup>[citation needed]</sup>, should we take the opportunity to nuke them completely like we did with v1 of XamlControlResources?

CC: @kazo0 @agneszitte

src/library/Uno.Toolkit.Material/Styles/Controls/v2/Chip.xaml
```diff
@@ -1,13 +1,11 @@
 ﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
 					xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
 					xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-					xmlns:not_win="http://uno.ui/not_win"
 					xmlns:todo="what should be done"
```
Xiaoy312, line 4: lets also do a sweep on the `todo` xmlns if unused (mc:Ignorable doesnt count) globally

src/library/Uno.Toolkit.Material/Styles/Controls/v2/NavigationBar.xaml
Xiaoy312, line 1: correct me if im wrong, but didnt we discontinue the support for winappsdk?
if so, we could also simplify those contractPresentOrNot xmlns

src/Uno.Toolkit.RuntimeTests/Helpers/XamlHelper.cs
```diff
@@ -27,7 +27,7 @@ internal static class XamlHelper
 		{
 			[string.Empty] = "http://schemas.microsoft.com/winfx/2006/xaml/presentation",
 			["x"] = "http://schemas.microsoft.com/winfx/2006/xaml",
-			["toolkit"] = "using:Uno.UI.Toolkit", // uno utilities
+			["toolkit"] = "using:Uno.UI.Xaml.Controls", // uno utilities (ex-Uno.UI.Toolkit; extensions now live in Uno.UI.Xaml)
```
Xiaoy312, line 30:
```suggestion
			["uuxc"] = "using:Uno.UI.Xaml.Controls", // uno utilities (ex-Uno.UI.Toolkit; extensions now live in Uno.UI.Xaml)
```
need pay attention to downstream consumers

src/Uno.Toolkit.UI/Extensions/DependencyObjectExtensions.cs
```diff
@@ -98,11 +99,7 @@ internal static IDisposable RegisterDisposablePropertyChangedCallback(this Depen
 		}
 
 		public static T? FindChild<T>(this DependencyObject depObj)
-		where T :
-#if HAS_UNO
-			class,
-#endif
-			DependencyObject
+		where T : DependencyObject
```
Xiaoy312, line 102:
```suggestion
			where T : DependencyObject
```

Xiaoy312 (approved):

src/library/Uno.Toolkit.Material/Styles/Controls/v1/_Common.xaml
agneszitte: I will let @kazo0 help validate this one

kazo0: We should drop v1 for this major release

samples/Directory.Packages.props
MartinZikmund: All four points are already in place on the current branch:

- #1636 merged on 2026-09-09 and this branch contains `origin/main` (`b1b8a2997`), so the `UnoThemesVersion` hunk and the `SimpleFontFamily` → `DefaultFontFamily` swap came in with the rebase rather than being resolved twice.
- `UnoThemesVersion` is `9.0.0-dev.15` in both `src/Directory.Packages.props` and `samples/Directory.Packages.props` — unoplatform/Uno.Themes#1722 published, so the blocker is cleared.
- The UTF-8 BOM on `samples/Directory.Packages.props` is gone (`head -c3` is `<Pr`).

Thanks for the heads-up on the `11.0-dev.{height}` / `servicing/10.0` split — this PR does not touch `version.json`, so no conflict there.

src/Uno.Toolkit.RuntimeTests/Tests/ScrollableHelperTests.cs
```diff
@@ -0,0 +1,61 @@
...
+	public async Task When_SmoothScrollTop_ListView_Scrolls_To_Start()
+	{
+		var (listView, scrollViewer) = await SetupListView();
+
+		scrollViewer.ChangeView(null, 500, null, disableAnimation: true);
+		await UnitTestUIContentHelperEx.WaitFor(() => scrollViewer.VerticalOffset > 0, message: "The ListView did not scroll down");
+
+		ScrollableHelper.SmoothScrollTop(listView);
+
+		await UnitTestUIContentHelperEx.WaitFor(() => scrollViewer.VerticalOffset == 0, timeoutMS: 3000, message: "The ListView was not scrolled back to the top");
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in `afccb54a4` — the wait predicate now uses a `TopOffsetTolerance` constant (0.5) instead of `VerticalOffset == 0`, so a smooth scroll settling on a sub-pixel offset no longer times the test out.

src/Uno.Toolkit.RuntimeTests/Tests/TabBarItemExtensionsTests.cs
```diff
@@ -0,0 +1,77 @@
...
+			?? contentHost.GetFirstDescendant<ScrollViewer>()
+			?? throw new AssertFailedException($"Failed to find the ScrollViewer of {contentHostType.Name}");
+
+		scrollViewer.ChangeView(null, 500, null, disableAnimation: true);
+		await UnitTestUIContentHelperEx.WaitFor(() => scrollViewer.VerticalOffset > 0, message: "The content host did not scroll down");
+		Assert.IsTrue(selectedItem.IsSelected);
+
+		selectedItem.ExecuteTap();
+
+		await UnitTestUIContentHelperEx.WaitFor(() => scrollViewer.VerticalOffset == 0, timeoutMS: 3000, message: "The content host was not scrolled back to the top");
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in `afccb54a4` — same treatment as `ScrollableHelperTests`: a `TopOffsetTolerance` constant (0.5) replaces the exact `VerticalOffset == 0` comparison.

src/Uno.Toolkit.UI/Controls/ExtendedSplashScreen/ExtendedSplashScreen.cross.cs
```diff
@@ -151,6 +194,93 @@ public partial class ExtendedSplashScreen
...
+					}
+
+					using var reader = new StreamReader(stream);
+					return ParseResizetizerDefinition(reader.ReadLine());
+				}
+			}
+			catch (Exception e)
+			{
+				typeof(ExtendedSplashScreen).Log().LogError(0, e, "Error while reading the splash screen definition.");
+			}
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in `afccb54a4`, though not quite as suggested — a `when` filter that only handles known types would let anything unexpected escape into app startup and take the window down, which is worse than what it replaces.

Instead it now follows `AGENTS.md` §8 (specific first, generic as the final fallback):

```csharp
catch (Exception e) when (e is IOException or BadImageFormatException or FileLoadException or NotSupportedException)
{
    typeof(ExtendedSplashScreen).Log().LogWarning(e, "Could not read the splash screen definition.");
}
catch (Exception e)
{
    typeof(ExtendedSplashScreen).Log().LogError(0, e, "Unexpected error while reading the splash screen definition.");
}
```

The foreseeable resource/stream failures are now a warning, anything else is still logged as an error, and either way the method degrades to `null` (no splash screen) rather than throwing.

src/library/Uno.Toolkit.Material/Styles/Controls/v1/BottomTabBar.xaml
MartinZikmund: Agreed — applied in `afccb54a4`, and swept globally rather than just here: all 10 style dictionaries plus the runtime-test `XamlHelper` xmlns map now use `uuxc`. No `toolkit` prefix is left in the repo.

The prefix is file-local in every case (all usages are `<uuxc:ElevatedView>` within the same dictionary), so nothing downstream binds to the old spelling.

src/Uno.Toolkit.RuntimeTests/Helpers/XamlHelper.cs
MartinZikmund: Applied in `afccb54a4`, along with the same rename in the 10 style dictionaries so the whole repo is consistent.

On downstream consumers: the prefix never escapes the file it is declared in — every usage is a `<uuxc:ElevatedView>` element inside the same dictionary, and `XamlHelper`'s map only feeds XAML strings built inside the runtime tests. Nothing external resolves `toolkit:`.

src/Uno.Toolkit.UI/Extensions/DependencyObjectExtensions.cs
MartinZikmund: Applied in `afccb54a4` — and to `GetFirstParent` too, since it had the same two-tab clause. `FindFirstParent` further down already used three tabs, so this just brings the file in line with itself.

src/library/Uno.Toolkit.Material/Styles/Controls/v2/Chip.xaml
MartinZikmund: Done in `afccb54a4`. `todo` was declared in 6 files and actually used in exactly one (`DrawerFlyoutSamplePage.xaml`, for a `todo:Note` attribute), so the other 5 declarations are gone — along with the `mc:Ignorable="todo"` entries and, where `todo` was the only thing `mc` carried, `xmlns:mc` itself.

The same pass removed the unreferenced `contract*Present` / `contract*NotPresent` / `primitiveContract7Present` declarations (see the `NavigationBar.xaml` thread).

Not swept: `d` and `local` in the sample pages. `d` is designer boilerplate and `local` is dead in ~60 sample files — a real clean-up, but pure churn to fold into a retarget PR. Happy to take it separately if you want it.

src/library/Uno.Toolkit.Material/Styles/Controls/v2/NavigationBar.xaml
MartinZikmund: Half right, so I dug into it — correcting the premise first.

**WinAppSDK is still supported at the library level.** `src/tfm-common-winui.props` still crosstargets `net10.0-windows10.0.19041` (gated by `Build_Windows`, which defaults to `true` on Windows hosts). What this PR dropped is the *sample heads'* Windows TFM: they are now `net10.0-android;net10.0-ios;net10.0-browserwasm;net10.0-desktop`. So consumers can still build against the WinAppSDK flavour of the packages.

**That said, the conditionals are still dead** — for a different reason:

- Uno's XAML generator resolves these at build time via `ApiInformation.IsApiContractPresent`, which reports `Windows.Foundation.UniversalApiContract` present **up to major 14** (`uno/src/Uno.Foundation/Metadata/ApiInformation.shared.cs`).
- On WinAppSDK the min target `10.0.19041` is UniversalApiContract **10**.

So `contract7Present` and `contract8Present` are always true on both, and `contract7NotPresent` is always false on both. Applied in `afccb54a4`:

- inlined `contract7Present:` (15 sites) and `contract8Present:` (23 sites) — all of them plain default-namespace attributes (`CornerRadius`, `VerticalAnchorRatio`) or elements (`ObjectAnimationUsingKeyFrames`, `DoubleAnimationUsingKeyFrames`, `DiscreteDoubleKeyFrame`, `SplineDoubleKeyFrame`, `VisualStateGroup.Transitions`), and none collided with an existing attribute on the same element;
- deleted the `contract7NotPresent:` branches (12 attributes + 4 element subtrees) that can never be taken;
- removed every `contract*Present` / `contract*NotPresent` / `primitiveContract7Present` declaration left unreferenced across `NavigationBar.xaml` (UI / Material v1 / Material v2 / Simple) and `DrawerControl.Enhanced.xaml`.

No `contract` prefix remains in any hand-written XAML. Desktop runtime tests are byte-identical before and after (430 tests, the same 91 display-scale-sensitive AutoLayout failures on a Windows host).

src/library/Uno.Toolkit.Material/Styles/Controls/v1/_Common.xaml
MartinZikmund: Agreed on all sides, but it lands separately — tracked in #1643 (assigned to @kazo0).

Confirming your citation: `v1` is not merely unused, it is already broken.

- `Themes/MaterialToolkitStyles.xaml` only merges `mergedpages.v2.xaml`, so nothing reaches `mergedpages.v1.xaml` through `MaterialToolkitTheme`.
- The one public way in, `MaterialToolkitResourcesV1`, points at `ms-appx:///Uno.Toolkit.WinUI.Material/Generated/mergedpages.WinUI.v1.xaml` — a path the merge task does not emit (the output is `mergedpages.v1.xaml`, no `.WinUI.` infix). Instantiating it throws.

`MaterialToolkitResourcesV2` has the identical broken `Source` URI, so #1643 covers that too.

Keeping the ~4,530-line deletion plus a public type removal out of this retarget PR — it is already large and green across every CI leg, and a public API removal deserves its own reviewable diff.

src/Uno.Toolkit.UI/Controls/ExtendedSplashScreen/ExtendedSplashScreen.cross.cs
```diff
@@ -151,6 +194,98 @@ public partial class ExtendedSplashScreen
...
+			catch (Exception e) when (e is IOException or BadImageFormatException or FileLoadException or NotSupportedException)
+			{
+				// The definition is embedded by Uno.Resizetizer; an unreadable assembly or stream just means no splash screen.
+				typeof(ExtendedSplashScreen).Log().LogWarning(e, "Could not read the splash screen definition.");
+			}
+			catch (Exception e)
+			{
+				typeof(ExtendedSplashScreen).Log().LogError(0, e, "Unexpected error while reading the splash screen definition.");
+			}
```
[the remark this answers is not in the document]
MartinZikmund: Declining this one, deliberately.

Deleting the fallback `catch (Exception e)` would let any unanticipated exception propagate out of `LoadSplashScreenFromResizetizerDefinition()`. That method runs during app startup while the splash screen is being resolved, so an escaping exception does not "improve diagnosability" — it takes the window down before the app is usable. On WASM in particular that is an unrecoverable crash in the worker, not a stack trace someone reads.

This also contradicts two rules in the repo's own `AGENTS.md`:

> §8: Order `catch` blocks from most specific to most general: catch predicted exceptions first […] **then use a generic `catch (Exception ex)` as the final fallback.**

> §8: For controls, prefer graceful degradation (e.g. fall back to a default visual state) over throwing from layout / template-application paths — exceptions in those paths can take down the visual tree on consumers.

The current shape is exactly what those prescribe, and it is what the previous round of this same check asked for (r4036584922): the foreseeable `IOException` / `BadImageFormatException` / `FileLoadException` / `NotSupportedException` are handled and logged as a warning, anything unexpected is logged at **error** level with the full exception — so it is not silently swallowed — and either path degrades to `null`, i.e. no splash screen.

Diagnosability is preserved by `LogError(0, e, …)`; app stability is preserved by not rethrowing. Keeping as is.

